PR #79, opened 2016-03-09
Fix mousetrap library path

kotcrab:
I fixed invalid `mousetrap` lib path in `gruntfile.js` (uppercase instead of lowercase). 
`grunt copy` was executing without errors even though `mousetrap` lib was not copied which resulted in blank page.


rr-:
Weird, I was almost sure I already fixed it yesterday

rr-:
...right, I committed it, but didn't push. :persevere: 

Thanks!

PR #81, opened 2016-04-05
Added arrow keys support

kotcrab:
Added support for navigation using arrow keys.


rr-:
Please update the text in the help / keyboard section as well. Could you also test how this behaves on Firefox and Chrome when creating a comment?

kotcrab:
Sure, I updated help and squashed commits but Github is having [some problems](https://status.github.com/messages) and it will take some time to appear here.

I tested it, hotkeys don't work when text field has focus. If they did then you wouldn't be able to type 'a', 'd' and others as well ;)

rr-:
...right, I'm being silly. I'll merge it tomorrow morning. Thanks!

PR #132, opened 2017-08-11
Fix build, use uglify-es package directly

mserajnik:
The `harmony`-branch is now its own package called `uglify-es`, which isn't backwards-compatible and breaks the build.

Using a non-ES6 version of `2.x` seems to work fine for now.

### DIFF
--- a/client/build.js
+++ b/client/build.js
@@ -70,7 +70,7 @@ function copyFile(source, target) {
 }
 
 function minifyJs(path) {
-    return require('uglify-js').minify(path, {compress: {unused: false}}).code;
+    return require('uglify-es').minify(fs.readFileSync(path, 'utf-8'), {compress: {unused: false}}).code;
 }
 
 function minifyCss(css) {
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,0 +1,2469 @@
+{
+  "name": "szurubooru",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "acorn": {
+      "version": "4.0.13",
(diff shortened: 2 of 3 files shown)


mserajnik:
Addition: I tried requiring `"uglify-es": "^3.0.0"` instead of `"uglify-js": "^2.0.0"` and changing the call in `build.js:73` to `return require('uglify-es').minify(path, {compress: {unused: false}}).code;` (which should still work the same according to the docs), but this results in `vendor.min.js` (and possibly also `app.min.js`, didn't check this one) just containing `undefined`.

rr-:
ES6 is necessary. ES5 may seem to work fine at first, but you'll see it breaks the site in random places.
I'll investigate exactly how the harmony branch has developed today.

mserajnik:
Thank you!

But aren't you transpiling with Babel beforehand and just using `uglify-js` for minifying? See `build.js:156` Because in that case it wouldn't matter if `uglify-js` can work with ES6, the code is ES5 already at that point in time.

Ah, it's because transpiling is optional, right? `uglify-js` would fail if `transpile: 1` is set in `config.yaml` because then it would actually have to deal with ES6 code.

rr-:
>Ah, it's because transpiling is optional, right? uglify-js would fail if transpile: 0 is set in config.yaml because then it would actually have to deal with ES6 code.

That is indeed the case.

mserajnik:
I seem to have fixed the issue – `uglify-es`'s `minify()` expects the actual code, not the path to the file containing the code.

Can you confirm that this works on your end too?

rr-:
Seems to work, thanks. Merged.

PR #133, opened 2017-08-12
Fix broken post fit modes on iOS 10 browsers

mserajnik:
With the way the post fit modes work, the image will always be displayed in full size on iOS 10 browsers, regardless of the setting.

I'm not actually sure if this is iOS 10 specific, but I think it might be because of https://twitter.com/thomasfuchs/status/742531231007559680/photo/1.

Setting a max-width on the body fixes it.

### DIFF
--- a/client/css/core-general.styl
+++ b/client/css/core-general.styl
@@ -13,6 +13,7 @@
 html
     height: 100%
 body
+    max-width: 100vw;
     min-height: 100%
 
 body


mserajnik:
After some further testing the issue might go a little bit deeper than I first thought. The fix works sometimes – opening a post, setting the post fit mode to _fit width_ and then refreshing, for example. In this case, the image will get the correct size.

However, going back to the posts overview and selecting the same post again, the size will be wrong again (even though the post fit mode didn't change). Refreshing then applies the correct size again.

rr-:
Please rebase the changes so I see only your commit

mserajnik:
Sorry, that went quite wrong. I actually rebased my develop branch on my master after merging upstream/master into that one because I forgot about this open PR. I reverted my develop branch to my initial commit now, should be fine again.

rr-:
This rule causes that large posts cannot be viewed at their original size because they're clipped.

Can you try tinkering with the following fragment in `post_main_view.js` instead?

```es6
            () => {
                return [
                    window.innerWidth -
                        postContainerNode.getBoundingClientRect().left -
                        margin,
                    window.innerHeight -
                        topNavigationNode.getBoundingClientRect().height -
                        margin * 2,
                ];
            });
```

mserajnik:
Yup, I've noticed that as well/didn't think about it in the beginning. And as I've said before, setting `max-width: 100vw;` on the `body` won't reliably fix the post fit modes issue either, so I think we'll really need to change the calculation. I haven't yet tested if this behavior has been fixed in iOS 11 but I guess that's very unlikely.

I'll try sorting this out when I find the time this or next week.

Aside from fixing this issue, I'd like to do some other mobile optimizations as well (maybe a drawer navigation instead of just floating the main navigation as it is now and also some iPhone X related improvements one could possibly apply to work around that notch in landscape mode – see http://stephenradford.me/removing-the-white-bars-in-safari-on-iphone-x/). I'll create separate PRs for those once I get my X and can test on that (I don't like using emulators for testing, too clunky).

mserajnik:
Okay, I've started looking into this issue some more. First, I can confirm it's still present on both Safari and Chrome on iOS 11 (but I'm pretty positive that both are using the exact same rendering engines since that's how Apple wants it on iOS).

When I just run

```javascript
console.log(
  JSON.stringify({
    'window.innerWidth': window.innerWidth,
    'window.innerHeight': window.innerHeight,
    'postContainerNode.getBoundingClientRect()': postContainerNode.getBoundingClientRect(),
    'topNavigationNode.getBoundingClientRect()': topNavigationNode.getBoundingClientRect()
  })
);
```

in the function you mentioned before returning I'm getting the following output in a case where it doesn't display the image correctly (on a real iPhone 7 running Chrome with fit mode _fit width_, I'm checking the console output via Safari on my Mac which allows me to open the developer tools for the USB-connected iPhone):

```json
{  
  "window.innerWidth":375,
  "window.innerHeight":628,
  "postContainerNode.getBoundingClientRect()":{  
    "x":5.625,
    "y":-57.375,
    "width":363.75,
    "height":0,
    "top":-57.375,
    "right":369.375,
    "bottom":-57.375,
    "left":5.625
  },
  "topNavigationNode.getBoundingClientRect()":{  
    "x":0,
    "y":-129,
    "width":375,
    "height":66,
    "top":-129,
    "right":375,
    "bottom":-63,
    "left":0
  }
}
```

When I then just refresh the page (after which it always works), the output looks like this:

```json
{  
  "window.innerWidth":375,
  "window.innerHeight":553,
  "postContainerNode.getBoundingClientRect()":{  
    "x":5.625,
    "y":71.625,
    "width":363.75,
    "height":0,
    "top":71.625,
    "right":369.375,
    "bottom":71.625,
    "left":5.625
  },
  "topNavigationNode.getBoundingClientRect()":{  
    "x":0,
    "y":0,
    "width":375,
    "height":66,
    "top":0,
    "right":375,
    "bottom":66,
    "left":0
  }
}
```

As you can see, there are some different height-related values. So I've thought about how this happens and noticed that it will always display it wrong when I open an image while the page is __scrolled down far enough in the post list that it causes the menu bar of the browser to be collapsed__ and always correctly when I open an image while the page is __scrolled up far enough in the post list that it causes the menu bar of the browser to be expanded__.

This is also why refreshing fixes it, it always scrolls back up to the top when doing so (and expands the menu bar of the browser).

I will have to look into it further, just thought I'd let you know in case you have any quick idea. If you run a search, there are tons of people posting on Stack Overflow regarding such issues with calculating heights and the usage of the `100vh` value in CSS on iOS browsers because of the changing menu bar height.

mserajnik:
The resizing issue mentioned here will be addressed with the new CSS-based resizing proposed in #149. I'll create a new PR since it's a broader change than just fixing the post fit modes on mobile.

PR #140, opened 2017-09-21
Add shortcut for deleting posts while in edit mode

icewind1991:
Add `del` as shortcut for deleting the post while the page is in edit mode


rr-:
Please add description for the new shortcut in https://github.com/rr-/szurubooru/blob/master/client/html/help_keyboard.tpl
otherwise looks good to me

icewind1991:
Done

rr-:
Merged manually
Thanks!

PR #141, opened 2017-09-21
preserve search query when using keyboard navigation

icewind1991:
Currently when using the keyboard to go the next/previous post, the search query is forgot after the first navigation.


rr-:
Merged manually

PR #145, opened 2017-12-15
Override resize mode in home view

### LINKED ISSUE #144: Automatic resizing for featured posts
It seems like the automatic resizing isn't working for featured posts. All elements with class `resize-listener` currently just get the original size of the image assigned. This causes the image to be displayed in full size regardless of the window size. No JS errors in the console.

featured_post_resizing

I assume the resizing on the frontpage should probably always behave like the fit mode "both" on post views? Or is this a missing feature rather than a bug?

mserajnik:
The featured post should always use the "both" resize mode in the home view regardless of what the user has set in the post view.

This resolves #144.

### DIFF
--- a/client/js/controls/post_content_control.js
+++ b/client/js/controls/post_content_control.js
@@ -5,18 +5,23 @@ const views = require('../util/views.js');
 const optimizedResize = require('../util/optimized_resize.js');
 
 class PostContentControl {
-    constructor(hostNode, post, viewportSizeCalculator) {
+    constructor(hostNode, post, viewportSizeCalculator, fitFunctionOverride) {
         this._post = post;
         this._viewportSizeCalculator = viewportSizeCalculator;
         this._hostNode = hostNode;
         this._template = views.getTemplate('post-content');
+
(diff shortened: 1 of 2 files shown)


client/js/controls/post_content_control.js
```diff
@@ -5,18 +5,23 @@ const views = require('../util/views.js');
...
-    constructor(hostNode, post, viewportSizeCalculator) {
+    constructor(hostNode, post, viewportSizeCalculator, fitFunctionOverride) {
         this._post = post;
         this._viewportSizeCalculator = viewportSizeCalculator;
         this._hostNode = hostNode;
         this._template = views.getTemplate('post-content');
 
+        let fitMode = settings.get().fitMode;
+        if (typeof fitFunctionOverride !== 'undefined') {
+            fitMode = this.fitBoth;
```
rr-: `fitMode = fitFunctionOverride`

mserajnik: Of course, I'll fix that right away. ^^

rr-:
Thanks :)

PR #146, opened 2017-12-16
Mobile styling improvements

mserajnik:
I've tried tweaking the mobile styles a little bit, namely:

- The top navigation now collapses
- The tags and tag categories tables no longer exceed the window width and are scrollable instead
- The search forms for posts, tags and users no longer exceed the window width and instead wrap their controls into a second line (or a third, if needed)
- The stats on the bottom of the home view no longer exceed the window width and instead display one item per line
- Buttons should no longer have weird gradient backgrounds and border radiuses on iOS browsers
- The comments list has reduced spacing between comments, I think it's still enough this way
- Some other little tweaks like the page counter not having an additional margin on the right for no apparent reason

I wasn't really sure what to do about the JS needed for the mobile navigation toggle, so I just put it somewhere where it's probably very wrong. From the looks of it, I guess you'd normally create a new file under `/control` for that? I'll have to look into that, but I figured I'd create the PR first so I can get some feedback.

Let me know what you think. There's certainly more stuff to improve, like the navigations when editing the account or a tag, I'll try to get to those as well.

Do you have any plans for a new frontend based on something like Vue? Not that this current one is bad or anything (I find it pretty impressive that you basically wrote a whole JS framework yourself), but I think I remember you mentioning that you'd like to do a rewrite in Vue/Angular/React/whatever before.

### DIFF
--- a/client/css/colors.styl
+++ b/client/css/colors.styl
@@ -55,3 +55,5 @@ $hovered-first-note-point-color = red
 $safety-safe = #88D488
 $safety-sketchy = #F3D75F
 $safety-unsafe = #F3985F
+$scrollbar-thumb-color = $main-color
+$scrollbar-bg-color = $input-enabled-background-color
--- a/client/css/comment-control.styl
+++ b/client/css/comment-control.styl
@@ -123,7 +123,7 @@ $comment-border-color = #DDD
         font-family: 'MS PGothic', 'ＭＳ Ｐゴシック', 'IPAMonaPGothic', 'Trebuchet MS', Verdana, Futura, Arial, Helvetica, sans-serif
         background: #fbfbfb
         color: #111
-        font-size: 12pt
+        font-size: 1em
         line-height: 1
         margin: 0
         padding: 4px
--- a/client/css/comment-control.styl
+++ b/client/css/comment-control.styl
@@ -3,7 +3,6 @@ $comment-header-background-color = $top-navigation-color
 $comment-border-color = #DDD
 
 .comment-container
-    margin: 0 0 1em 0
     padding: 0 0 0 60px
 
     .avatar
--- a/client/css/comment-list-control.styl
+++ b/client/css/comment-list-control.styl
@@ -2,3 +2,8 @@
     list-style-type: none
     margin: 0
     padding: 0
+
+    >li
+        margin-bottom: 1em
+        &:last-child
+            margin-bottom: 0
--- a/client/css/comment-list-view.styl
+++ b/client/css/comment-list-view.styl
@@ -1,15 +1,24 @@
+@import colors
+$comment-border-color = $top-navigation-color
+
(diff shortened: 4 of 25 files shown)


mserajnik:
I've added some more little tweaks (mainly stuff that I wanted to make consistent but where I overlooked something) – I've amended the original commit a few times.

I've also added a few new commits which might be more controversial: the first one changes all `pt` font size declarations to `em`. I feel like it's both more in line with the rest of the styling (where most declarations are done in `em`) and also easier to read. Was there a specific reason to use `pt` over `em` or even `px`? Nothing particularly wrong with it, but I don't see it used very much these days, that's why I'm asking.

I've also added default font sizes for `h1` and `h2` where previously only the default browser styles would apply. They are the same sizes as the Chrome default so that should be fine.

The last change I did was changing the paddings of `#content-holder` and `.content-wrapper` to use `em` instead of `vw`. Basically the same reasoning as with the font sizes, more in line with the rest. In this case it also makes sense if you think about different screen sizes. Someone with a screen with lots of real estate (let's say a UWQHD monitor) might want to use the built-in zoom functionality of the browser. While everything styled with `em` or percentages will now get bigger proportionally to each other, everything styled with `vw` will still be the same size due to the browser still calculating from the "real" base viewport width. Doesn't look too good and also doesn't make much sense in my opinion.

Again, let me know what you think, those new commits are more suggestions based on personal preference, feel free to discard them if you don't like them.

mserajnik:
Done with the tab navigations as well. I have nothing more to add for now. Did some testing and everything looks good to me.

rr-:
https://github.com/rr-/szurubooru/pull/146/commits/758a291c4299d4f7d012ef2f6115e6326e43405b

Could you make these margins a little bigger on desktop?

mserajnik:
Done, I've increased it some, let me know what you think. d36ff49

I don't want to go to big on the `#content-holder` padding since I'd like the navigation items to have the same padding for alignment. Setting it too big would cause the navigation to break on smaller screens (before displaying the collapsed version).

mserajnik:
I've just looked through these changes again and I can't find anything else I'd like to tweak at the moment. Do you want me to change anything else or do you think this can be merged?

rr-:
Sorry, forgot about this. I'm gonna review it within a few days.

mserajnik:
Not worries, I actually kinda forgot about it myself until today. :D No rush, either.

rr-:
I haven't had time to review this yet, but I'll try to find some this weekend.

rr-:
Rebased and merged:

- a98ca553
- db913243
- 23a28ce6
- a962bb35
- a08c7d65
- 7596f904
- 9b10d2be
- e15dffa1
- 4ce29cf2
- 26a1451f

Great changes. I like hamburger menu and scrollable navigation very much. Thanks :)

PR #154, opened 2018-02-28
Hardening and Quality of life changes

ReAnzu:
### List of proposed changes:
- Allow other users to register new users based on users:create:any permission. Default permission is administrator.
- Add 2 new configuration options to allow automatic conversion of animated gif to mp4 and webm during post creation. Default configuration is off.
- When ffmpeg errors out during thumbnail creation, but some byte output is generated, use that byte output for the thumbnail. ffmpeg has a tendency to error when generating thumbnails from improperly formatted mp4 and webm sources, but still generates a correct thumbnail.
- When posts are deleted, delete the source file and the generated thumbnail.
- Harden the user password hashing to use libsodium python bindings and the argon2id hashing function.
- Automatically convert legacy password hashes to the new password hash on successful password based login.
- Add token based authentication so that the client doesn't need to store plaintext user credentials in the cookie to support the remember me functionality. Or on the api to handle authenticating subsequent requests.
- Add user_token API endpoints to list, create, update, and delete user tokens.
- Add permissions for the user_token API endpoints. Defaults to regular for self, and administrator for any.
- Change the client to use token based authentication after the initial login.A new token is requested after successful login bump and upon logout the token is deleted.
- Remove the plaintext user password instances stored in the API and site auth cookie.
- Add a UI for managing tokens as a tabbed view on the user UI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ config.yaml
 */*_modules/
 .coverage
 .cache
+__pycache__
+.idea/
+*.iml
+data/
--- a/API.md
+++ b/API.md
@@ -7,6 +7,7 @@
 1. [General rules](#general-rules)
 
    - [Authentication](#authentication)
+   - [User token authentication](#user-token-authentication)
    - [Basic requests](#basic-requests)
    - [File uploads](#file-uploads)
    - [Error handling](#error-handling)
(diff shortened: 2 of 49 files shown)


rr-:
No. Split this into multiple pull requests.

ReAnzu:
Sure, I can do that.

I see probably 4-5 discrete pull requests in here.

PR #155, opened 2018-03-01
Introduce users:create:any permission and update UI to allow creation of users manually

ReAnzu:
users:create:any permission and UI update
* Added functionality for a user to directly add users to the application. This allows for registration to be closed for anonymous, but still have users created, by authorized users.
* Added permission users:create:any to handle level that users are allowed to create other users
* Moved old permission users:create to users:create:self

### DIFF
--- a/server/szurubooru/api/user_api.py
+++ b/server/szurubooru/api/user_api.py
@@ -45,9 +45,6 @@ def create_user(
     ctx.session.add(user)
     ctx.session.commit()
 
-    if ctx.user.user_id is not None:
-        user = ctx.user
-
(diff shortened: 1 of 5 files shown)


server/szurubooru/api/user_api.py
```diff
@@ -40,6 +44,10 @@ def create_user(
             ctx.get_file('avatar', default=b''))
     ctx.session.add(user)
     ctx.session.commit()
+
+    if ctx.user.user_id is not None:
+        user = ctx.user
```
rr-: Wait, why do we return logged in user to the logged in user?

rr-:
Merged

PR #157, opened 2018-03-01
Allow thumbnails to be generated even if ffmpeg throws warnings

ReAnzu:
* Poorly formatted MP4 and WEBM sources can cause ffmpeg to throw a lot of warnings. However when there is byte ouptut, the generated thumbnail is valid. Add a bypass for the resize_fill function to allow ffmpeg to error.
* Simplify the ffmpeg video filter scaling parameter

### DIFF
--- a/server/szurubooru/func/images.py
+++ b/server/szurubooru/func/images.py
@@ -33,10 +33,14 @@ class Image:
         return self.info['streams'][0]['nb_read_frames']
 
     def resize_fill(self, width: int, height: int) -> None:
+
+        width_greater = self.width > self.height
+        scale_param = "scale='%d:%d" % ((-1, height) if width_greater else (width, -1))
+
         cli = [
             '-i', '{path}',
             '-f', 'image2',
-            '-vf', _SCALE_FIT_FMT.format(width=width, height=height),
+            '-filter:v', scale_param,
             '-map', '0:v:0',
             '-vframes', '1',
             '-vcodec', 'png',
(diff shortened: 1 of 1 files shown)


rr-:
If the new filter works, I'd remove `_SCALE_FIT_FMT` constant.

server/szurubooru/func/images.py
```diff
@@ -33,10 +33,14 @@ def frames(self) -> int:
         return self.info['streams'][0]['nb_read_frames']
 
     def resize_fill(self, width: int, height: int) -> None:
+
```
rr-: Unneeded blank line

ReAnzu:
Looking back at the code, I think the .format looks cleaner so I'll adapt the code to fit the original style.

server/szurubooru/func/images.py
```diff
@@ -12,7 +12,7 @@
 
 
 _SCALE_FIT_FMT = (
```
rr-: Now it's simple enough that it can be just inlined

server/szurubooru/func/images.py
```diff
@@ -79,7 +82,8 @@ def to_jpeg(self) -> bytes:
             '-',
         ])
 
-    def _execute(self, cli: List[str], program: str = 'ffmpeg') -> bytes:
+    def _execute(self, cli: List[str], program: str = 'ffmpeg',
+                 ignore_error_if_data: bool = False) -> bytes:
```
rr-: Indentation

PR #158, opened 2018-03-01
Added auto conversion options to the config for gif to mp4 and webm

ReAnzu:
* webm conversion is slow, but better quality than mp4 conversion with a typically smaller file size
* tags are copied over from the original upload
* Snapshots are generated for the new auto posts
* Default setting is false for both conversions, as this will require additional resources of the server, but is bandwidth friendly for viewers

### DIFF
--- a/server/szurubooru/api/post_api.py
+++ b/server/szurubooru/api/post_api.py
@@ -72,15 +72,18 @@ def create_post(
     create_snapshots_for_post(post, new_tags, None if anonymous else ctx.user)
     alternate_format_posts = posts.generate_alternate_formats(post, content)
     for alternate_post, alternate_post_new_tags in alternate_format_posts:
-        create_snapshots_for_post(alternate_post,
-                                  alternate_post_new_tags,
-                                  None if anonymous else ctx.user)
+        create_snapshots_for_post(
+            alternate_post,
+            alternate_post_new_tags,
+            None if anonymous else ctx.user)
     ctx.session.commit()
     return _serialize_post(ctx, post)
 
 
-def create_snapshots_for_post(post: model.Post, new_tags: List[model.Tag],
-                              user: Optional[model.User]):
+def create_snapshots_for_post(
+        post: model.Post,
+        new_tags: List[model.Tag],
+        user: Optional[model.User]):
     snapshots.create(post, user)
     for tag in new_tags:
         snapshots.create(tag, user)
--- a/server/szurubooru/api/post_api.py
+++ b/server/szurubooru/api/post_api.py
@@ -72,12 +72,15 @@ def create_post(
     create_snapshots_for_post(post, new_tags, None if anonymous else ctx.user)
     alternate_format_posts = posts.generate_alternate_formats(post, content)
     for alternate_post, alternate_post_new_tags in alternate_format_posts:
-        create_snapshots_for_post(alternate_post, alternate_post_new_tags, None if anonymous else ctx.user)
+        create_snapshots_for_post(alternate_post,
+                                  alternate_post_new_tags,
+                                  None if anonymous else ctx.user)
     ctx.session.commit()
     return _serialize_post(ctx, post)
 
 
-def create_snapshots_for_post(post: model.Post, new_tags: List[model.Tag], user: Optional[model.User]):
+def create_snapshots_for_post(post: model.Post, new_tags: List[model.Tag],
+                              user: Optional[model.User]):
     snapshots.create(post, user)
     for tag in new_tags:
         snapshots.create(tag, user)
--- a/server/szurubooru/api/post_api.py
+++ b/server/szurubooru/api/post_api.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict
+from typing import Optional, Dict, List
 from datetime import datetime
 from szurubooru import db, model, errors, rest, search
 from szurubooru.func import (
@@ -69,11 +69,18 @@ def create_post(
         posts.update_post_thumbnail(post, ctx.get_file('thumbnail'))
     ctx.session.add(post)
     ctx.session.flush()
-    snapshots.create(post, None if anonymous else ctx.user)
-    for tag in new_tags:
-        snapshots.create(tag, None if anonymous else ctx.user)
+    create_snapshots_for_post(post, new_tags, None if anonymous else ctx.user)
+    alternate_format_posts = posts.generate_alternate_formats(post, content)
+    for alternate_post, alternate_post_new_tags in alternate_format_posts:
+        create_snapshots_for_post(alternate_post, alternate_post_new_tags, None if anonymous else ctx.user)
     ctx.session.commit()
     return _serialize_post(ctx, post)
+
+
+def create_snapshots_for_post(post: model.Post, new_tags: List[model.Tag], user: Optional[model.User]):
+    snapshots.create(post, user)
+    for tag in new_tags:
+        snapshots.create(tag, user)
 
 
 @rest.routes.get('/post/(?P<post_id>[^/]+)/?')
--- a/server/szurubooru/func/images.py
+++ b/server/szurubooru/func/images.py
@@ -134,9 +134,7 @@ class Image:
             ]
 
             if altered_dimensions:
-                args = args + [
-                    '-filter:v', 'scale=\'%d:%d\'' % (width, height)
-                ]
+                args += ['-filter:v', 'scale=\'%d:%d\'' % (width, height)]
 
             self._execute(args + ['-y', mp4_temp_path])
 
--- a/server/szurubooru/func/images.py
+++ b/server/szurubooru/func/images.py
@@ -79,6 +79,72 @@ class Image:
             '-',
         ])
 
+    def to_webm(self) -> bytes:
+        with util.create_temp_file_path(suffix='.log') as phase_log_path:
+            # Pass 1
+            self._execute([
+                '-i', '{path}',
+                '-pass', '1',
+                '-passlogfile', phase_log_path,
+                '-vcodec', 'libvpx-vp9',
+                '-crf', '4',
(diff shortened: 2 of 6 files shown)


server/szurubooru/func/images.py
```diff
@@ -79,6 +79,72 @@ def to_jpeg(self) -> bytes:
...
+                '-vcodec', 'libvpx-vp9',
+                '-crf', '4',
+                '-b:v', '2500K',
+                '-acodec', 'libvorbis',
+                '-f', 'webm',
+                '-'
+            ])
+
+    def to_mp4(self) -> bytes:
+
```
rr-: Extra blank line

server/szurubooru/func/images.py
```diff
@@ -79,6 +79,72 @@ def to_jpeg(self) -> bytes:
...
+        with util.create_temp_file_path(suffix='.dat') as mp4_temp_path:
+
```
rr-: And another

server/szurubooru/func/mime.py
```diff
@@ -1,58 +1,66 @@
 import re
 from typing import Optional
 
+APPLICATION_SWF = 'application/x-shockwave-flash'
+IMAGE_JPEG = 'image/jpeg'
+IMAGE_PNG = 'image/png'
+IMAGE_GIF = 'image/gif'
+VIDEO_WEBM = 'video/webm'
+VIDEO_MP4 = 'video/mp4'
+APPLICATION_OCTET_STREAM = 'application/octet-stream'
```
rr-: Don't we exagerrate a little bit?

server/szurubooru/func/mime.py
```diff
@@ -1,58 +1,66 @@
...
 
 
 def is_flash(mime_type: str) -> bool:
-    return mime_type.lower() == 'application/x-shockwave-flash'
+    return mime_type.lower() == APPLICATION_SWF
 
 
 def is_video(mime_type: str) -> bool:
-    return mime_type.lower() in ('application/ogg', 'video/mp4', 'video/webm')
+    return mime_type.lower() in ('application/ogg', VIDEO_MP4, VIDEO_WEBM)
```
rr-: And now we keep `ogg` as magic string…?

server/szurubooru/func/util.py
```diff
@@ -41,6 +41,16 @@ def create_temp_file(**kwargs: Any) -> Generator:
         os.remove(path)
 
 
+@contextmanager
+def create_temp_file_path(**kwargs: Any) -> Generator:
+    (descriptor, path) = tempfile.mkstemp(**kwargs)
+    os.close(descriptor)
```
rr-: Hey that's pretty cool function.

rr-:
So the gist of it is that you just generate new posts and let the users decide afterwards which version they want to keep, correct?

What if the backend converts GIF to WEBM on upload rather than creating duplicate posts?

ReAnzu:
You're right, there is the flexibility to delete the original animation in favor of the auto generated mp4 or webm. However sometimes the generation might go poorly (it took me a few iterations on a previous project to come up with the ffmpeg settings for webm, and I'm still not happy with the performance hit a 2 pass encode requires). So I'm not sure if not persisting the gif is ideal.

Also some people probably prefer gif even if there is a bandwidth advantage to a video format.

Perhaps an ideal feature would be to allow a post to have multiple formats and then let the administrator determine the system preferred formats to serve by default with extra formats available via an API call? This would mean that auto conversion could be added for png to jpeg, or non animated gif to png, etc.

server/szurubooru/func/util.py
ReAnzu: Ah, well I can't take credit for that. It's 90% of the function above it, I just needed the file path instead of an open file descriptor.

rr-:
>Perhaps an ideal feature would be to allow a post to have multiple formats and then let the administrator determine the system preferred formats to serve by default with extra formats available via an API call? This would mean that auto conversion could be added for png to jpeg, or non animated gif to png, etc.

Hm, yes, but that makes things a little bit bloated IMO. You'd need to handle this in:

- UI
    - Post view
    - Post upload
    - Post edit
- API
    - Upload
    - Edit
    - Get
    - Decision which file to use to make a thumbnail
    - Duplicate upload detection
    - Similar upload detection (the part that queries elasticsearch)
- Snapshots

While that would be a cool feature to have, the effort to benefit ratio seems to be too high, so I'd rather go with your approach.

server/szurubooru/api/post_api.py
```diff
@@ -69,13 +69,23 @@ def create_post(
...
     ctx.session.flush()
-    snapshots.create(post, None if anonymous else ctx.user)
-    for tag in new_tags:
-        snapshots.create(tag, None if anonymous else ctx.user)
+    create_snapshots_for_post(post, new_tags, None if anonymous else ctx.user)
+    alternate_format_posts = posts.generate_alternate_formats(post, content)
+    for alternate_post, alternate_post_new_tags in alternate_format_posts:
+        create_snapshots_for_post(alternate_post,
+                                  alternate_post_new_tags,
+                                  None if anonymous else ctx.user)
```
rr-: Indentation

server/szurubooru/api/post_api.py
```diff
@@ -69,13 +69,23 @@ def create_post(
...
     ctx.session.commit()
     return _serialize_post(ctx, post)
 
 
+def create_snapshots_for_post(post: model.Post, new_tags: List[model.Tag],
+                              user: Optional[model.User]):
```
rr-: Indentation

server/szurubooru/func/images.py
```diff
@@ -79,6 +79,70 @@ def to_jpeg(self) -> bytes:
...
+                '-crf', '22',
+                '-b:v', '200K',
+                '-profile:v', 'main',
+                '-pix_fmt', 'yuv420p',
+                '-acodec', 'aac',
+                '-f', 'mp4'
+            ]
+
+            if altered_dimensions:
+                args = args + [
```
rr-: `args += [` ?

server/szurubooru/func/mime.py
```diff
@@ -1,58 +1,67 @@
...
 from typing import Optional
 
+_APPLICATION_OCTET_STREAM = 'application/octet-stream'
+_APPLICATION_OGG = 'application/ogg'
+_APPLICATION_SWF = 'application/x-shockwave-flash'
+_IMAGE_GIF = 'image/gif'
+_IMAGE_JPEG = 'image/jpeg'
+_IMAGE_PNG = 'image/png'
+_VIDEO_MP4 = 'video/mp4'
+_VIDEO_WEBM = 'video/webm'
```
rr-: I still don't see benefit of these TBH

server/szurubooru/func/mime.py
```diff
@@ -1,58 +1,67 @@
...
-    return mime_type.lower() in ('image/jpeg', 'image/png', 'image/gif')
+    return mime_type.lower() in (_IMAGE_JPEG, _IMAGE_PNG, _IMAGE_GIF)
 
 
 def is_animated_gif(content: bytes) -> bool:
     pattern = b'\x21\xF9\x04[\x00-\xFF]{4}\x00[\x2C\x21]'
-    return get_mime_type(content) == 'image/gif' \
-        and len(re.findall(pattern, content)) > 1
+    return (get_mime_type(content) == _IMAGE_GIF
+            and len(re.findall(pattern, content)) > 1)
```
rr-: Indentation

server/szurubooru/func/posts.py
```diff
@@ -429,6 +429,42 @@ def _sync_post_content(post: model.Post) -> None:
...
 
 
+def generate_alternate_formats(post: model.Post, content: bytes) \
+        -> List[Tuple[model.Post, List[model.Tag]]]:
+    assert post
+    assert content
+    new_posts = []
+    if mime.is_animated_gif(content):
+        tag_names = [tag_name.name for tag_name in
+                     [tag.names for tag in post.tags]]
```
rr-: Indentation

        tag_names = [
            tag_name.name
            for tag_name in [
                 tag.names
                 for tag in post.tags]]

or

        tag_names = [
            tag_name.name
            for tag_name in [tag.names for tag in post.tags]]

server/szurubooru/func/posts.py
```diff
@@ -429,6 +429,42 @@ def _sync_post_content(post: model.Post) -> None:
...
+
+        if config.config['convert']['gif']['to_mp4']:
+            mp4_post, new_tags = create_post(images.Image(content).to_mp4(),
+                                             tag_names, post.user)
```
rr-: Indentation

server/szurubooru/func/posts.py
```diff
@@ -429,6 +429,42 @@ def _sync_post_content(post: model.Post) -> None:
...
+            update_post_flags(mp4_post, ['loop'])
+            update_post_safety(mp4_post, post.safety)
+            update_post_source(mp4_post, post.source)
+            new_posts += [(mp4_post, new_tags)]
+
+        if config.config['convert']['gif']['to_webm']:
+            webm_post, new_tags = create_post(images.Image(content).to_webm(),
+                                              tag_names, post.user)
```
rr-: Indentation

server/szurubooru/func/posts.py
```diff
@@ -429,6 +429,42 @@ def _sync_post_content(post: model.Post) -> None:
...
+            update_post_flags(webm_post, ['loop'])
+            update_post_safety(webm_post, post.safety)
+            update_post_source(webm_post, post.source)
+            new_posts += [(webm_post, new_tags)]
+
+        db.session.flush()
+
+        new_posts = list(filter(lambda i: i[0] is not None, new_posts))
```
rr-: `new_posts = [p for p in new_posts if p[0] is not None]` ?

PR #159, opened 2018-03-01
Harden Password Hashes

ReAnzu:
* Changed password setup to use libsodium and argon2id
* Regular SHA256 hashing for passwords is inadequate as modern GPU's can hash generate billions of hashes per second.
* Added code to auto migrate old passwords to the new password_hash if the existing password_hash matches either of the legacy password generation schemes (SHA1 or SHA256).
* Added migration to support new password_hash format length
* Added auth tests


rr-:
That's what salt is for, billions per second is still nothing if you have a good salt.

server/szurubooru/func/auth.py
```diff
@@ -1,8 +1,11 @@
 import hashlib
 import random
 from collections import OrderedDict
-from szurubooru import config, model, errors
+from nacl.exceptions import InvalidkeyError
+
+from szurubooru import config, model, errors, db
 from szurubooru.func import util
+from nacl.pwhash import argon2id, verify
```
rr-: I'd rather use `from nacl import pwhash` rather than importing functions with such bogus names as `verify`

server/szurubooru/func/auth.py
```diff
@@ -47,11 +57,22 @@ def create_password() -> str:
...
+        return verify(user.password_hash.encode('utf8'),
+                      (config.config['secret'] + salt + password).encode('utf8'))
+    except InvalidkeyError:
+        possible_hashes = [
+            get_sha256_legacy_password_hash(salt, password),
+            get_sha1_legacy_password_hash(salt, password)
+        ]
+        if valid_hash in possible_hashes:
+            # Convert the user password hash to the new hash
+            user.password_hash = get_password_hash(salt, password)
```
rr-: That's a great idea, I haven't thought of this before.
Could you add a column that determines version of the password hash?
That way DBA could have a rough idea how many legacy hashes are in use.

server/szurubooru/migrations/versions/9ef1a1643c2a_increase_password_hash_max_field_length.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+revision = '9ef1a1643c2a'
+down_revision = '02ef5f73f4ab'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
```
rr-: Unneeded blank line

server/szurubooru/migrations/versions/9ef1a1643c2a_increase_password_hash_max_field_length.py
```diff
@@ -0,0 +1,30 @@
...
+    op.alter_column('user', 'password_hash',
+               existing_type=sa.VARCHAR(length=64),
```
rr-: Weird indentation

rr-:
Other than requested changes, LGTM

ReAnzu:
I'll cleanup the commented sections.

I think adding a password_revision field should do the trick. Determining password hash revision should be easy at the moment, SHA1 and SHA256 have different fixed widths, so that data could be populated during the migration.

ReAnzu:
In regards to salt, it certainly thwarts attack multiplication against many users with the same actual password. But if the hashes are being calculated that means the attacker has the database or a dump of the user table and they'll have the salts for each user. Then it's just the server secret that obfuscates the hash. But if the server is compromised, then the secret might be compromised as well. (granted if the server is compromised, it might be easier for an attacker to just shim the authentication code and dump the passwords as plain text...)

Which means that the attacker probably knows the secret and the salt for a specific user. Because the code is open, they know the the order that the secret and salt are appended to the password. Which means it's just down to how long it will take to permute over the remaining password possibilities. SHA256 is really fast to compute, so it wouldn't take long to match particular passwords, with only one round of SHA255 encoding.

Granted all of this is a just a delaying tactic to slow down the actual cracking of the password.

Ultimately it just made me nervous, as I don't want something I run to potentially leak approximately easy to break credentials.

rr-:
Well it's like you said: if you have salt and server secret and the code, you can loop through the user list and compare their hashes to top 10 passwords from a list of the most popular passwords all the same, it'll just take a little bit longer.

That said I don't oppose using stronger hash, of course :P

ReAnzu:
Okay, there is now a password_revision field on the user table. 1 corresponds to old SHA1 hashes as determined by their characteristic 40 character width. 2 corresponds to the SHA256 hashes, as determined by their 64 character width. 3 is reserved for the argon2id hashes, but those aren't part of the remediation in the migration as they won't have been saved yet.

Any hash that isn't length 40 or 64 will be set to password_revision 0.

I think that rounds out this feature branch.

server/szurubooru/func/auth.py
```diff
@@ -1,8 +1,13 @@
...
+
 import hashlib
 import random
 from collections import OrderedDict
-from szurubooru import config, model, errors
+from nacl.exceptions import InvalidkeyError
+
+from szurubooru import config, model, errors, db
 from szurubooru.func import util
+from nacl import pwhash
```
rr-: Imports

server/szurubooru/func/users.py
```diff
@@ -243,7 +243,9 @@ def update_user_password(user: model.User, password: str) -> None:
         raise InvalidPasswordError(
             'Password must satisfy regex %r.' % password_regex)
     user.password_salt = auth.create_password()
-    user.password_hash = auth.get_password_hash(user.password_salt, password)
+    hash, revision = auth.get_password_hash(user.password_salt, password)
```
rr-: Keyword shadowing, linters gonna complain

server/szurubooru/func/users.py
```diff
@@ -308,5 +310,7 @@ def reset_user_password(user: model.User) -> str:
     assert user
     password = auth.create_password()
     user.password_salt = auth.create_password()
-    user.password_hash = auth.get_password_hash(user.password_salt, password)
+    hash, revision = auth.get_password_hash(user.password_salt, password)
```
rr-: Keyword shadowing

server/szurubooru/migrations/env.py
```diff
@@ -35,7 +35,11 @@ def run_migrations_offline():
     '''
     url = alembic_config.get_main_option('sqlalchemy.url')
     alembic.context.configure(
-        url=url, target_metadata=target_metadata, literal_binds=True)
+        url=url,
+        target_metadata=target_metadata,
+        literal_binds=True,
+        compare_type=True
+    )
```
rr-: Hanging `)` can be merged to the line above

server/szurubooru/migrations/env.py
```diff
@@ -56,7 +60,9 @@ def run_migrations_online():
     with connectable.connect() as connection:
         alembic.context.configure(
             connection=connection,
-            target_metadata=target_metadata)
+            target_metadata=target_metadata,
+            compare_type=True
+        )
```
rr-: Hanging `)` can be merged to the line above

server/szurubooru/migrations/versions/9ef1a1643c2a_update_user_table_for_hardened_passwords.py
```diff
@@ -0,0 +1,81 @@
...
+
+    user_id = sa.Column('id', sa.Integer, primary_key=True)
+    creation_time = sa.Column('creation_time', sa.DateTime, nullable=False)
+    last_login_time = sa.Column('last_login_time', sa.DateTime)
+    version = sa.Column('version', sa.Integer, default=1, nullable=False)
+    name = sa.Column('name', sa.Unicode(50), nullable=False, unique=True)
+    password_hash = sa.Column('password_hash', sa.Unicode(128), nullable=False)
+    password_salt = sa.Column('password_salt', sa.Unicode(32))
+    password_revision = sa.Column('password_revision', sa.SmallInteger,
+                                  default=0, nullable=False)
```
rr-: Indentation

server/szurubooru/migrations/versions/9ef1a1643c2a_update_user_table_for_hardened_passwords.py
```diff
@@ -0,0 +1,81 @@
...
+        'version_id_col': version,
+        'version_id_generator': False,
+    }
+
+
+def upgrade():
+    op.alter_column('user', 'password_hash',
+                    existing_type=sa.VARCHAR(length=64),
+                    type_=sa.Unicode(length=128),
+                    existing_nullable=False)
```
rr-: Indentation

server/szurubooru/migrations/versions/9ef1a1643c2a_update_user_table_for_hardened_passwords.py
```diff
@@ -0,0 +1,81 @@
...
+    op.add_column('user', sa.Column('password_revision',
+                                    sa.SmallInteger(),
+                                    nullable=True,
+                                    default=0))
```
rr-: Indentation

server/szurubooru/migrations/versions/9ef1a1643c2a_update_user_table_for_hardened_passwords.py
```diff
@@ -0,0 +1,81 @@
...
+                user.password_revision = 1
+            elif password_hash_length == 64:
+                user.password_revision = 2
+            else:
+                user.password_revision = 0
+        session.flush()
+    session.commit()
+
+    op.alter_column('user', 'password_revision',
+                    existing_nullable=True, nullable=False)
```
rr-: Indentation

server/szurubooru/migrations/versions/9ef1a1643c2a_update_user_table_for_hardened_passwords.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def downgrade():
+    op.alter_column('user', 'password_hash',
+                    existing_type=sa.Unicode(length=128),
+                    type_=sa.VARCHAR(length=64),
+                    existing_nullable=False)
```
rr-: Indentation

server/szurubooru/model/user.py
```diff
@@ -23,8 +23,10 @@ class User(Base):
     last_login_time = sa.Column('last_login_time', sa.DateTime)
     version = sa.Column('version', sa.Integer, default=1, nullable=False)
     name = sa.Column('name', sa.Unicode(50), nullable=False, unique=True)
-    password_hash = sa.Column('password_hash', sa.Unicode(64), nullable=False)
+    password_hash = sa.Column('password_hash', sa.Unicode(128), nullable=False)
     password_salt = sa.Column('password_salt', sa.Unicode(32))
+    password_revision = sa.Column('password_revision', sa.SmallInteger,
+                                  default=0, nullable=False)
```
rr-: Indentation

server/szurubooru/tests/conftest.py
```diff
@@ -115,11 +115,15 @@ def injector(new_config_content):
 
 @pytest.fixture
 def user_factory():
-    def factory(name=None, rank=model.User.RANK_REGULAR, email='dummy'):
+    def factory(name=None,
+                rank=model.User.RANK_REGULAR,
+                email='dummy',
+                password_salt=None,
+                password_hash=None):
```
rr-: Indentation

rr-:
Not sure why GH doesn't show this up here but one last thing:

https://github.com/rr-/szurubooru/pull/159/files#r172998061

rr-:
Squashed into https://github.com/rr-/szurubooru/commit/3f52aceca44bc52e0c8654f46c66073320109ae9

PR #160, opened 2018-03-01
User Token Authentication

ReAnzu:
* Users are only authenticated against their password on login, and to retrieve a token.
* Passwords are wiped from the app and cookies after login and token retrieval
* Tokens are revoked at the end of the session/logout
* If the user chooses the "remember me" option, the token is stored in the cookie
* A user interface to revoke tokens will be added
* Tokens correctly delete themselves on logout
* API documentation updated for the new user-token endpoints
* Added a Manage tokens tab to the user panel
* Added bullet point about the token authentication for the API
* Added tests for new endpoints and tests against authentication middleware

* The migration file in this commit was developed after the password hardening feature. So if these are merged in other orders or if a new migration is deployed before this is merged, the migration hashes will need to be updated.


.gitignore
```diff
@@ -2,3 +2,6 @@ config.yaml
 */*_modules/
 .coverage
 .cache
+__pycache__
+.idea/
+*.iml
```
rr-: No

client/js/api.js
```diff
@@ -87,16 +88,16 @@ class Api extends events.EventTarget {
...
         const auth = cookies.getJSON('auth');
-        return auth && auth.user && auth.password ?
-            this.login(auth.user, auth.password, true) :
+        return auth && auth.user && auth.token ?
+            this.login_with_token(auth.user, auth.token, true) :
             Promise.resolve();
     }
 
-    login(userName, userPassword, doRemember) {
+    login_with_token(userName, token, doRemember) {
```
rr-: pascalCase

client/js/api.js
```diff
@@ -105,8 +106,63 @@ class Api extends events.EventTarget {
...
+                    this.userName = userName;
+                    this.userToken = response.token;
+                    this.userPassword = null;
+                }, error => {
+                    reject(error);
+                });
+        });
+    }
+
+    delete_token(userName, userToken) {
```
rr-: pascalCase

client/js/controllers/user_controller.js
```diff
@@ -21,8 +22,11 @@ class UserController {
             return;
         }
 
+        this._successMessages = [];
+        this._errorMessages = [];
+
         topNavigation.setTitle('User ' + userName);
-        User.get(userName).then(user => {
+        User.get(userName).then(async user => {
```
rr-: `async`?

client/js/controllers/user_controller.js
```diff
@@ -58,18 +73,50 @@ class UserController {
...
                 ranks: ranks,
+                tokens: userTokens,
             });
             this._view.addEventListener('change', e => this._evtChange(e));
             this._view.addEventListener('submit', e => this._evtUpdate(e));
             this._view.addEventListener('delete', e => this._evtDelete(e));
+            this._view.addEventListener('create-token', e => this._evtCreateToken(e));
+            this._view.addEventListener('delete-token', e => this._evtDeleteToken(e));
+
+            for (let i = 0; i < this._successMessages.length; i++) {
```
rr-: Weren't there better ways of iterating in ES6?

client/js/models/user_token.js
```diff
@@ -0,0 +1,76 @@
...
+const events = require('../events.js');
+
+class UserToken extends events.EventTarget {
+    constructor() {
+        super();
+        this._orig = {};
+        this._updateFromResponse({});
+    }
+
+    get token()               { return this._token; }
```
rr-: Indentation

client/js/models/user_token.js
```diff
@@ -0,0 +1,76 @@
...
+    get enabled()               { return this._enabled; }
```
rr-: _pascalCase (applies to all other fields)

client/js/views/user_tokens_view.js
```diff
@@ -0,0 +1,86 @@
...
+        views.decorateValidator(this._formNode);
+
+        this._formNode.addEventListener('submit', e => this._evtSubmit(e));
+
+        this._decorateTokenForms()
+    }
+
+    _decorateTokenForms() {
+        for (let i = 0; i < this._tokens.length; i++) {
+            let formNode = this._hostNode.querySelector('#token' + i);
```
rr-: Change `id="token{i}"` to `class="token"` and add new attribute `data-token-id="i"` in the HTML template, then change the selector to `.token[data-token-id=i]`

client/js/views/user_tokens_view.js
```diff
@@ -0,0 +1,86 @@
...
+        views.disableForm(this._formNode);
+        for (let i = 0; i < this._tokenFormNodes.length; i++) {
+            let formNode = this._tokenFormNodes[i];
+            views.disableForm(formNode);
+        }
+    }
+
+    _evtDelete(e) {
+        e.preventDefault();
+        const userToken = this._tokens[parseInt(e.target.id.replace('token', ''))];
```
rr-: `parseInt(e.target.getAttribute('data-token-id'))

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,86 @@
...
+        user_token: Optional[model.UserToken],
+        auth_user: model.User,
+        options: List[str] = []) -> Optional[rest.Response]:
+    if not user_token:
+        return None
+    return UserTokenSerializer(user_token, auth_user).serialize(options)
+
+
+def get_user_token_by_user_and_token(user: model.User, token: str) -> model.UserToken:
+    return (db.session.query(model.UserToken)
```
rr-: Indentation

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,86 @@
...
+            .filter(model.UserToken.user_id == user.user_id, model.UserToken.token == token)
+            .one_or_none())
+
+
+def get_user_tokens(user: model.User) -> List[model.UserToken]:
+    assert user
+    return (db.session.query(model.UserToken)
```
rr-: Indentation

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,86 @@
...
+    user_token.enabled = True
+    user_token.creation_time = datetime.utcnow()
+    db.session.add(user_token)
+    db.session.commit()
+    return user_token
+
+
+def update_user_token_enabled(user_token: model.UserToken, enabled: bool) -> None:
+    assert user_token
+    user_token.enabled = enabled if enabled is not None else True
```
rr-: If `enabled` is `bool` then it can't be `None`, can it?

server/szurubooru/func/users.py
```diff
@@ -172,9 +174,9 @@ def get_user_count() -> int:
 def try_get_user_by_name(name: str) -> Optional[model.User]:
     return (
         db.session
-        .query(model.User)
-        .filter(sa.func.lower(model.User.name) == sa.func.lower(name))
-        .one_or_none())
+            .query(model.User)
```
rr-: `pycodestyle` is gonna complain about that kind of indentation IIRC

server/szurubooru/middleware/authenticator.py
```diff
@@ -13,19 +13,33 @@ def _authenticate(username: str, password: str) -> model.User:
     return user
 
 
+def _authenticate_token(username: str, token: str) -> model.User:
+    """Try to authenticate user. Throw AuthError for invalid users."""
```
rr-: Use `'''` if possible

server/szurubooru/middleware/authenticator.py
```diff
@@ -13,19 +13,33 @@ def _authenticate(username: str, password: str) -> model.User:
...
+        else:
             raise HttpBadRequest(
                 'ValidationError',
                 'Only basic HTTP authentication is supported.')
-        username, password = base64.decodebytes(
-            credentials.encode('ascii')).decode('utf8').split(':', 1)
-        return _authenticate(username, password)
     except ValueError as err:
         msg = (
             'Basic authentication header value are not properly formed. '
```
rr-: Now the error message can also apply to token authentication. The message should be adjusted to apply to both authentication schemes

server/szurubooru/migrations/versions/a39c7f98a7fa_add_user_token_table.py
```diff
@@ -0,0 +1,35 @@
...
+
+revision = 'a39c7f98a7fa'
+down_revision = '9ef1a1643c2a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table('user_token',
+    sa.Column('id', sa.Integer(), nullable=False),
```
rr-: Indentation

server/szurubooru/model/__init__.py
```diff
@@ -1,5 +1,7 @@
 from szurubooru.model.base import Base
-from szurubooru.model.user import User
+from szurubooru.model.user import (
+    User,
```
rr-: Nah, I'd say keep it in one line

server/szurubooru/tests/api/test_user_token_creating.py
```diff
@@ -0,0 +1,29 @@
...
+    with patch('szurubooru.func.user_tokens.create_user_token'), \
+            patch('szurubooru.func.user_tokens.serialize_user_token'), \
+            patch('szurubooru.func.users.get_user_by_name'), \
+            fake_datetime('1969-02-12'):
+        users.get_user_by_name.return_value = user_token.user
+        user_tokens.serialize_user_token.return_value = 'serialized user token'
+        user_tokens.create_user_token.return_value = user_token
+        result = api.user_token_api.create_user_token(
+            context_factory(
+                user=user_token.user),
```
rr-: No need for the indentation, same goes to other tests

rr-:
Rad changes, I'll have to take a look at the UI later on.

ReAnzu:
Thanks for the review, I'll address the comments and cleanup the code.

I'm also thinking now that I've used it a bit on my dev instance, that a note and expiration field would be nice to have for the tokens.

client/js/controllers/user_controller.js
ReAnzu: Ah, that's because I added an await to the internal call for getting the tokens for the user if we're on that view. It needs to happen synchronously before the view is created for that page.

server/szurubooru/func/user_tokens.py
ReAnzu: For some reason I don't think github registered the indentation change

rr-:
Thanks. I'll finish reviewing this PR after we merge the previous ones.

ReAnzu:
Thanks for taking the time to review, my python styling is not particularly strong. I'm sure I'll be making some styling changes to these commits after I cleanup the other pull requests.

client/js/controllers/user_controller.js
ReAnzu: I'll need to test this with transpiling back to ES5, I'm not sure if that will work. If it doesn't I'll need to work around that.

client/js/api.js
```diff
@@ -258,7 +333,12 @@ class Api extends events.EventTarget {
             }
 
             try {
-                if (this.userName && this.userPassword) {
+                if (this.userName && this.userToken) {
+                    req.auth = null;
+                    req.set('Authorization', 'Token '
+                        + new Buffer(this.userName + ":" + this.userToken).toString('base64'))
+                }
+                else if (this.userName && this.userPassword) {
```
rr-: `} else if {`

client/js/models/user_token.js
```diff
@@ -0,0 +1,89 @@
...
+            .then(response => {
+                return Promise.resolve(UserToken.fromResponse(response));
+            });
+    }
+
+    static create(userName, note, expirationTime) {
+        let userTokenRequest = {
+            enabled: true
+        };
+        if (note){
```
rr-: Missing space

.gitignore
```diff
@@ -1,4 +1,4 @@
 config.yaml
 */*_modules/
 .coverage
-.cache
+.cache
```
rr-: No real changes to the file, I'd reset it

client/js/models/user_token.js
```diff
@@ -0,0 +1,89 @@
...
+            _version:        response.version,
+            _creationTime:   response.creationTime,
+        };
+
+        Object.assign(this, map);
+        Object.assign(this._orig, map);
+    }
+}
+
+module.exports = UserToken;
```
rr-: Missing `\n` at the end of the file

(motivation: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline )

client/js/util/date.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+Date.prototype.addDays = function(days) {
+    let dat = new Date(this.valueOf());
+    dat.setDate(dat.getDate() + days);
+    return dat;
+};
```
rr-: Missing `\n` at the end of the file

client/js/views/user_tokens_view.js
```diff
@@ -0,0 +1,102 @@
...
+        e.preventDefault();
+        this.dispatchEvent(new CustomEvent('submit', {
+            detail: {
+                user: this._user,
+
+                note: this._userTokenNoteInputNode ?
+                    this._userTokenNoteInputNode.value :
+                    undefined,
+
+                expirationTime: this._userTokenExpirationTimeInputNode && this._userTokenExpirationTimeInputNode.value.length > 0 ?
```
rr-: I'd wrap this line

server/szurubooru/api/user_token_api.py
```diff
@@ -0,0 +1,84 @@
+from typing import Dict
+
```
rr-: Unnecessary newline

server/szurubooru/api/user_token_api.py
```diff
@@ -0,0 +1,84 @@
...
+    auth.verify_privilege(ctx.user, 'user_tokens:create:%s' % infix)
+    enabled = ctx.get_param_as_bool('enabled', True)
+    user_token = user_tokens.create_user_token(user, enabled)
+    if ctx.has_param('note'):
+        note = ctx.get_param_as_string('note')
+        user_tokens.update_user_token_note(user_token, note)
+    if ctx.has_param('expirationTime'):
+        expiration_time = ctx.get_param_as_string('expirationTime')
+        user_tokens.update_user_token_expiration_time(user_token,
+                                                      expiration_time)
```
rr-: Indentation

server/szurubooru/api/user_token_api.py
```diff
@@ -0,0 +1,84 @@
...
+    user = users.get_user_by_name(params['user_name'])
+    infix = 'self' if ctx.user.user_id == user.user_id else 'any'
+    auth.verify_privilege(ctx.user, 'user_tokens:edit:%s' % infix)
+    user_token = user_tokens.get_by_user_and_token(user, params['user_token'])
+    versions.verify_version(user_token, ctx)
+    versions.bump_version(user_token)
+    if ctx.has_param('enabled'):
+        auth.verify_privilege(ctx.user, 'user_tokens:edit:%s' % infix)
+        user_tokens.update_user_token_enabled(user_token,
+                                              ctx.get_param_as_bool('enabled'))
```
rr-: Indentation

server/szurubooru/api/user_token_api.py
```diff
@@ -0,0 +1,84 @@
...
+    if ctx.has_param('note'):
+        auth.verify_privilege(ctx.user, 'user_tokens:edit:%s' % infix)
+        note = ctx.get_param_as_string('note')
+        user_tokens.update_user_token_note(user_token, note)
+    if ctx.has_param('expirationTime'):
+        auth.verify_privilege(ctx.user, 'user_tokens:edit:%s' % infix)
+        expiration_time = ctx.get_param_as_string('expirationTime')
+        user_tokens.update_user_token_expiration_time(user_token,
+                                                      expiration_time)
```
rr-: Indentation

server/szurubooru/func/auth.py
```diff
@@ -54,6 +56,18 @@ def is_valid_password(user: model.User, password: str) -> bool:
     return valid_hash in possible_hashes
 
 
+def is_valid_token(user_token: model.UserToken) -> bool:
+    ''' Token must be enabled and if it has an expiration,
```
rr-: Indentation

```
'''
Lorem ipsum dolor sit amet
'''
```

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,138 @@
...
+        return None
+    return UserTokenSerializer(user_token, auth_user).serialize(options)
+
+
+def get_by_user_and_token(
+        user: model.User, token: str) -> model.UserToken:
+    return (db.session.query(model.UserToken)
+            .filter(model.UserToken.user_id == user.user_id,
+                    model.UserToken.token == token)
+            .one_or_none())
```
rr-: Indentation

```
    return (
        db.session
            .query(model.UserToken)
            .filter(
                model.UserToken.user_id == user.user_id,
                model.UserToken.token == token)
            .one_or_none())
```

or sth like that

server/szurubooru/func/user_tokens.py
rr-:
```
return (
    db.session.query(model.UserToken)
    .filter(...)
    .all())
```

or sth like that

server/szurubooru/func/users.py
```diff
@@ -170,11 +170,10 @@ def get_user_count() -> int:
...
 def try_get_user_by_name(name: str) -> Optional[model.User]:
-    return (
-        db.session
-        .query(model.User)
-        .filter(sa.func.lower(model.User.name) == sa.func.lower(name))
-        .one_or_none())
+    return (db.session
+            .query(model.User)
+            .filter(sa.func.lower(model.User.name) == sa.func.lower(name))
+            .one_or_none())
```
rr-: Indentation

server/szurubooru/middleware/authenticator.py
```diff
@@ -13,29 +13,42 @@ def _authenticate(username: str, password: str) -> model.User:
...
     if not ctx.has_header('Authorization'):
         return None
 
     try:
         auth_type, credentials = ctx.get_header('Authorization').split(' ', 1)
-        if auth_type.lower() != 'basic':
+        if auth_type.lower() == 'basic':
+            username, password = base64.decodebytes(
+                credentials.encode('ascii')).decode('utf8').split(':', 1)
+            return _authenticate(username, password)
```
rr-: Rename to `_authenticate_basic_auth`

server/szurubooru/migrations/versions/a39c7f98a7fa_add_user_token_table.py
```diff
@@ -0,0 +1,38 @@
...
+                    sa.Column('token', sa.Unicode(length=36), nullable=False),
+                    sa.Column('note', sa.Unicode(length=128), nullable=True),
+                    sa.Column('enabled', sa.Boolean(), nullable=False),
+                    sa.Column('expiration_time', sa.DateTime(), nullable=True),
+                    sa.Column('creation_time', sa.DateTime(), nullable=False),
+                    sa.Column('last_edit_time', sa.DateTime(), nullable=True),
+                    sa.Column('version', sa.Integer(), nullable=False),
+                    sa.ForeignKeyConstraint(
+                        ['user_id'], ['user.id'], ondelete='CASCADE'),
+                    sa.PrimaryKeyConstraint('id'))
```
rr-: Indentation

server/szurubooru/tests/api/test_user_token_updating.py
```diff
@@ -0,0 +1,40 @@
...
+            patch('szurubooru.func.users.get_user_by_name'), \
+            fake_datetime('1969-02-12'):
+        users.get_user_by_name.return_value = user_token.user
+        user_tokens.serialize_user_token.return_value = 'serialized user token'
+        user_tokens.get_by_user_and_token.return_value = user_token
+        result = api.user_token_api.update_user_token(
+            context_factory(params={'version': user_token.version,
+                                    'enabled': False,
+                                    },
+                            user=user_token.user),
```
rr-: Indentation

server/szurubooru/tests/middleware/test_authenticator.py
```diff
@@ -0,0 +1,51 @@
+from unittest.mock import patch
+
```
rr-: Unneccessary blank line

server/szurubooru/tests/func/test_user_tokens.py
```diff
@@ -0,0 +1,75 @@
...
+    assert result is None
+
+
+def test_get_by_user_and_token(user_token_factory):
+    user_token = user_token_factory()
+    db.session.add(user_token)
+    db.session.flush()
+    db.session.commit()
+    result = user_tokens.get_by_user_and_token(user_token.user,
+                                               user_token.token)
```
rr-: Indentation

server/szurubooru/tests/func/test_user_tokens.py
```diff
@@ -0,0 +1,75 @@
+from datetime import datetime
+from unittest.mock import patch
+
```
rr-: Unnecessary blank line

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,142 @@
+import pytz
```
rr-: Doesn't this need to go to `requirements.txt` as well?

client/js/controllers/user_controller.js
rr-: In case you wind up needing to change it, here you can see how I load models into views.

First, the controller asks for the model:

```
User.get(id)
```

This returns a promise which I can then react to:

```
User.get(id).then(
    user => { success! let's construct the view },
    error => { show the error somewhere... })
```

In some places I also await for multiple resources at once:

```
Promise.all([
    TagCategoryList.get(),
    Tag.get(ctx.parameters.name),
]).then(responses => {
    const [tagCategoriesResponse, tag] = responses;
```

rr-: I didn't know about `async` and now my promise-based code looks super-ugly. Oh well...

server/szurubooru/func/user_tokens.py
ReAnzu: Yes probably.

What's odd is that, I've recently rebuilt my virtualenv with 3.5 since I encountered some odd 3.6 v 3.5 issues related to datetime handling and I didn't explicitly install a pytz version. Perhaps it is coming in as a transient dependency from another library. Still, it should be an explicit library requirement.

client/js/controllers/user_controller.js
ReAnzu: It looks like async/await doesn't transpile backwards from es6 to es5 (without a babel plugin), and support has only been common since late 2016 / early 2017 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) So this may be a little new to rely on.

Given that I'll be adapting this to use the Promise.all design. It might not be prettier, but it will be consistent with the codebase and won't increase requirements for the users browser.

server/szurubooru/func/user_tokens.py
ReAnzu: I've added it to the requirements.txt

rr-:
Testing the UI and found following bug:

1. Login
2. Token gets created
3. Delete the token
4. `Internal Server Error`

```
[2018-03-10 18:50:43] waitress Exception when serving /user-tokens/eax
Traceback (most recent call last):
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/channel.py", line 338, in servic
    task.service()
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/task.py", line 399, in execute
    app_iter = self.channel.server.application(env, start_response)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/rest/app.py", line 97, in application
    hook(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 64, in process_request_hook
    process_request(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 54, in process_request
    auth_user = _get_user(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 38, in _get_user
    return _authenticate_token(username, token)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 20, in _authenticate_token
    if not auth.is_valid_token(user_token):
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/func/auth.py", line 89, in is_valid_token
    assert user_token
AssertionError
[2018-03-10 18:50:43] waitress Exception when serving /user/eax
Traceback (most recent call last):
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/channel.py", line 338, in servic
    task.service()
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/home/rr-/src/maintained/szurubooru/server/python_modules/lib/python3.6/site-packages/waitress/task.py", line 399, in execute
    app_iter = self.channel.server.application(env, start_response)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/rest/app.py", line 97, in application
    hook(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 64, in process_request_hook
    process_request(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 54, in process_request
    auth_user = _get_user(ctx)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 38, in _get_user
    return _authenticate_token(username, token)
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/middleware/authenticator.py", line 20, in _authenticate_token
    if not auth.is_valid_token(user_token):
  File "/home/rr-/src/maintained/szurubooru/server/szurubooru/func/auth.py", line 89, in is_valid_token
    assert user_token
AssertionError
```

server/szurubooru/tests/func/test_auth.py
```diff
@@ -1,7 +1,5 @@
-from datetime import datetime, timedelta
 import pytest
```
rr-: pytest is a custom pkg, importing datetime and timedelta should go before

rr-:
Additionally, the list of tokens could show each token's expiration date.

ReAnzu:
Agreed on the UI, I've updated the flow to logout the user if deleting the same token than the client is using for authentication. For clarity, should that delete button instead have a text of logout and a description explaining why?

That assert on the token shouldn't have been there. As it makes sense that if no tokens exist, a None might be passed to the method. In those cases a None is most certainly not a valid token, so the authentication should fail, not throw an exception.

rr-:
>For clarity, should that delete button instead have a text of logout and a description explaining why?

That would be cool, perhaps `Delete and logout` and the explanation would go to the tooltip.

rr-:
Regarding UI:

1. I'd make the box a little narrow like it used to be
2. See if aligning the text to the left (but keeping it in columns) looks better
3. If there's no space for the delete button to appear on the right, move it to the bottom like this:

    ```
    Token:    0de863d1-2456-4d77-83f3-616875d28ac4
    Note:     Client Login Token
    Created:  3/10/2018
    Expires:  No Expiration

    [￼Delete token]
    ```

4. Dates should use `views.makeRelativeTime()` (you can still see the exact day in the tooltip)

ReAnzu:
image

- Returned the width to 35em. It does cause the overflow to kick in on the menus. Before the mobile tweaks the Account deletion would flow to the next line.
- Used an hr instead of a custom border between tokens. I did change the border-top attribute to make it slightly darker than the default #ddd
- I'm not sure I'm a fan of putting the user agent in the note by default on login. Long user agents might overflow the column width in the API, which is currently set at 128 characters (My Chrome user agent is 105 characters long, and my Firefox user agent is 76 characters long). Also I'm not sure it would add much to the user experience to know the browser user agent that created the token.
- I adjusted the form input to match the other input dialogs in the user section. I started by making it look like another token row, and while it looked nice it was inconsistent with the rest of the user tabs. Also it required some forcing on minimum height to make the labels align, so it may not have been particularly responsive to screen size differences.

rr-:
What do you think about:

```patch
diff --git a/client/html/user.tpl b/client/html/user.tpl
index cbf4f18f..75c721f3 100644
--- a/client/html/user.tpl
+++ b/client/html/user.tpl
@@ -4,13 +4,13 @@
         --><ul><!--
             --><li data-name='summary'><a href='<%- ctx.formatClientLink('user', ctx.user.name) %>'>Summary</a></li><!--
             --><% if (ctx.canEditAnything) { %><!--
-                --><li data-name='edit'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'edit') %>'>Account settings</a></li><!--
+                --><li data-name='edit'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'edit') %>'>Settings</a></li><!--
             --><% } %><!--
             --><% if (ctx.canListTokens) { %><!--
-                --><li data-name='list-tokens'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'list-tokens') %>'>Manage tokens</a></li><!--
+                --><li data-name='list-tokens'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'list-tokens') %>'>Login tokens</a></li><!--
             --><% } %><!--
             --><% if (ctx.canDelete) { %><!--
-                --><li data-name='delete'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'delete') %>'>Account deletion</a></li><!--
+                --><li data-name='delete'><a href='<%- ctx.formatClientLink('user', ctx.user.name, 'delete') %>'>Delete</a></li><!--
             --><% } %><!--
         --></ul><!--
     --></nav>
```

(By the way, with this change, the delete button becomes consistent with the tag edit view.)

---

Also, what do you think about adding last usage date to the user token? (that'd use `?bump-login` to avoid executing DB `UPDATE` every API query)

---

Finally, could we have UI to edit token names? Rather than reworking the whole thing or creating complicated subforms, I'd suggest just a small `(change)` link near each name, which would ask with `prompt()` for new name and execute AJAX immediately, showing errors through either `showError` or `alert` - it doesn't need to be fancy since it will be used only once in a blue moon.

rr-:
I'd also update `<hr/>` to make it consistent with global comment list:

```patch
diff --git a/client/css/user-view.styl b/client/css/user-view.styl
index 3640cac9..8389202b 100644
--- a/client/css/user-view.styl
+++ b/client/css/user-view.styl
@@ -1,3 +1,6 @@
+@import colors
+$token-border-color = $active-tab-background-color
+
 #user
     width: 100%
     max-width: 35em
@@ -76,9 +79,7 @@
                 padding-right: 0.5em
 
             hr
-                border-top: 1px #aaa solid;
+                border-top: 3px solid $token-border-color
 
     #user-delete form
         width: 100%
-
-
```

rr-:
Also, rather than typing `<input` please use `make*Input` that also adds relevant CSS classes. (If you look at the inputs in the token form you'll notice that they have 1px border rather than 3px border.)

Additionally, please use `attribute='value'` rather than `attribute="value"` for consistency.

ReAnzu:
Sure, I'll address these. I've also merged those patches.

- The bump-login change seems simple enough, I'll add a new field and integrate it with the existing flow.
- By edit name, you mean the note field right? Adding a prompt to change the note seems reasonable, and I don't see why the event couldn't propagate any error back to the page.
- I'll work on cleaning up that view, now that the style seems settled, to try and match what I see done in other parts of the UI.

server/szurubooru/func/user_tokens.py
```diff
@@ -0,0 +1,152 @@
...
+from datetime import datetime
+from pyrfc3339 import parser as rfc3339_parser
+from typing import Any, Optional, List, Dict, Callable
+from szurubooru import db, model, rest, errors
+from szurubooru.func import auth, serialization, users, util
+
+
+class InvalidEnabledError(errors.ValidationError):
```
tehoko: This exception doesn't seem to have tests.

rr-:
Actually last night I've pushed some commit and now it seems to be gone. Please don't use `-f` from now on

ReAnzu:
I am very sorry. I was rebasing elsewhere and must've left the flag on my command line before I pushed that last set of changes.

server/szurubooru/func/user_tokens.py
ReAnzu: Hmm, that exception shouldn't be there. enabled can't be None because it is coming from ctx.get_param_as_bool().

I've removed the exception and also the guard code in the update method.

API.md
```diff
@@ -56,6 +57,11 @@
         - [Updating user](#updating-user)
         - [Getting user](#getting-user)
         - [Deleting user](#deleting-user)
+    - User Tokens
+        - [Listing tokens](#listing-tokens)
```
mserajnik: This link and the three following after this one are currently broken because the headings are named differently. Based on their names it should be:

+ `Listing user tokens` => `#listing-user-tokens`
+ `Creating a user token` => `#creating-a-user-token`
+ `Updating a user token` => `#updating-a-user-token`
+ `Deleting a user token` => `#deleting-a-user-token`

rr-:
I think it's good to merge, is there anything you would like to add?

ReAnzu:
I think the additional changes look good, and I haven't encountered any further issues on my local development server. I think the feature is good to merge.

rr-:
Rebased and merged

PR #161, opened 2018-03-07
Add example config for the Caddy HTTP server

itskenny0:


CrawlerBleak:
I saw this issue on Google and needed a config for Caddy v2. Here's what I came up with:

```
booru.example {
	redir /api /api/
	route /api/* {
		uri strip_prefix /api
		header * Access-Control-Allow-Origin *

		reverse_proxy unix//var/run/booru/booru.sock	
	}
	route * {
		root * /var/web/booru
		try_files {path} /index.htm
		file_server {
			index index.htm index.html
		}
	}
	route /media/* {
		root * /var/lib/booru
		uri strip_prefix /media
		file_server
	}
}
```

It doesn't have all the same error handling or gzip support as the nginx config, but it seems to work. You may also want to modify the `reverse_proxy` line.

Issue #153, opened 2018-02-13
"Parameter ‘tags’ must be a list of string values" when copying tags from existing post


TehTotalPwnage:
I uploaded two versions of the same picture onto my personal booru setup and the uploader warned me about the two images having a very high similarity. I opted to add a relation between the two images and to copy the tags from the old image to the new one, but when pressing upload, the uploader states `One of the posts needs your attention; click “resume upload” when you’re ready.` with the error on the image being `Parameter ‘tags’ must be a list of string values.`.

I've managed to reproduce this with other images, which seems to suggest that the issue isn't limited to just that image. Aside from that, I can't think of much else that might be the cause of the issue.

ReAnzu:
It looks like the issue was in the client sending all of the tag data, when the API expected only a list of tag names.

rr-:
Fix merged in HEAD. Thanks

PR #163, opened 2018-03-09
client/posts: fix copy tags list of string values error #153

ReAnzu:
Issue #153 seems to be because the API expects a list of tag names, but the client is sending the entire list of tag objects when Copy tag is selected on the upload dialog for the similar images. This resulted in ctx.get_param_as_string_list method iterating over objects that weren't strings and returning the "Parameter ‘tags’ must be a list of string values" error message.

I think the simplest fix is to add a get method to the post object to retrieve a list of the tag names. Referencing only the first name in the list of names for a tag seems to be the current practice for identifying the primary name for a tag.

### DIFF
--- a/client/js/models/post.js
+++ b/client/js/models/post.js
@@ -39,6 +39,7 @@ class Post extends events.EventTarget {
 
     get flags()              { return this._flags; }
     get tags()               { return this._tags; }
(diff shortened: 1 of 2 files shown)


rr-:
Thanks

PR #164, opened 2018-04-05
Fix post fit calculation on iOS browsers

### LINKED ISSUE #149: Proposal: use CSS-based resizing
The current JavaScript-based resizing has some issues, especially on mobile/iOS devices (see #133).

There are also sporadic hiccups in desktop browsers where it calculates the width or height a few pixels off (mostly when resizing the viewport) resulting in imaginary lines drawn between the elements not aligning and the view looking off.

Therefore, I'd like to propose replacing the JS-based solution with a pure CSS solution. I've prepared a simple mockup that should give an idea of how this could work: https://codepen.io/mserajnik/full/ypaMZV/

This solution would be viable for the post view where the medium can take the full height of the viewport minus the navigation bar (which could simply be calculated using `calc()`).

It's a little more difficult in the home view where there are text elements which can of course be dynamic in height. A possible solution for that would be to lay out the vertical structure in the home view with Flexbox and give the container containing the medium the `flex: 1` property to use the remaining available space (of course, there should still be a minimum height even if that causes scrolling on very small screens like landscape phones). This would make it possible to use a form of the `both` mode there as well (as it's currently the case too, although not working perfectly, especially on smaller screens).

Browser support shouldn't need to be considered since I've not used any CSS features more advanced than what's already in use somewhere; therefore, the requirements wouldn't change.

mserajnik:
This should fix the issue I've originally mentioned in #133. It also closes #149 (at least for now, maybe the CSS-based resizing is still something we could add in the future, just for performances sake).

Other browsers shouldn't be affected because the helper library I'm using for calculating a correct viewport height on iOS (independent of the status bar) checks the user agent and just returns `window.innerHeight` (like it was before) in case anything else than iOS is detected.

I've changed the calculation of `margin` because we can't rely on `postViewNode.getBoundingClientRect().top` on iOS, it returns different values depending on the state of the status bar. `sidebarNode.getBoundingClientRect().left` on the other hand stays the same (because it is independent of the available viewport height). Additionally, moving its calculation inside the function passed in the `PostContentControl()` instantiation ensures it gets updated correctly when resizing the viewport and hitting a media query breakpoint where the padding gets smaller/larger.

### DIFF
--- a/client/js/views/post_main_view.js
+++ b/client/js/views/post_main_view.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const iosCorrectedInnerHeight = require('ios-inner-height');
 const router = require('../router.js');
 const views = require('../util/views.js');
 const uri = require('../util/uri.js');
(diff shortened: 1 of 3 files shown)


client/js/views/post_main_view.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const iosCorrectedInnerHeight = require('ios-inner-height');
```
mserajnik: I've corrected the variable name from `iOSCorrectedInnerHeight`.

rr-:
Merged: https://github.com/rr-/szurubooru/commit/d39439d549a003880933d730f80c56eaf2cf20fa

PR #166, opened 2018-05-21
client/posts: fix upload error caused by anonymous node

rishubil:
Anonymous node does not exist in view when a user without anonymous upload permission tries to post upload. So in this case we should check for the existence of anonymousNode first.


rr-:
Thanks!

Issue #168, opened 2018-06-22
Allow for fetching of client side configs during runtime?


sgsunder:
Is there any reason why some of the client-side configurations have to be built beforehand, instead of being fetched from the server at runtime?

Specifically, the privileges and password, username, and tag regexes in the config.yaml file could be fetched from the server as needed, instead of being baked into a JSON file at buildtime. I also noticed that `info_api.py` on the server side already provides this functionality.

The reason I ask this is because being able to hold off on providing the config.yaml file until runtime will make it much easier to run within Docker. I'm already using a Docker deployment and it works quite well and plays nicely with official postgres and elasticsearch images, however the fact that I can't run `node build.js` until after I've passed a config file in makes orchestration really clunky.

rr-:
There's no real reason other than the info API being an afterthought (hence the technical debt you pointed out). If you wish to change this behavior, I encourage you to fire a PR.

PR #170, opened 2018-06-24
Get client-side configuration at runtime

sgsunder:
Per #168 

Most of the needed client-side configurations, including the privileges table, regular expression filters, app title, and contact info, are fetched using the backend's Info API now. 

The hope is that this will greatly simplify the build/deployment process, especially for Docker users, as it will be possible to build the frontend before many of the configurations are set.

I have a modified `build.js` that I'm working on, but I'm excluding it from this PR (for now) as it would alter the build/deployment process for existing users. Right now, this PR should work as a drop-in change and end users won't have to change anything about their deployment process.

### DIFF
--- a/client/js/api.js
+++ b/client/js/api.js
@@ -8,6 +8,7 @@ const progress = require('./util/progress.js');
 const uri = require('./util/uri.js');
 
 let fileTokens = {};
+let remoteConfig = null;
 
 class Api extends events.EventTarget {
     constructor() {
@@ -65,14 +66,53 @@ class Api extends events.EventTarget {
         return this._wrappedRequest(url, request.delete, data, {}, options);
     }
 
+    fetchConfig() {
+        if (remoteConfig === null) {
+            return this.get(uri.formatApiLink('info'))
+                .then(response => {
+                    remoteConfig = response.config;
+                });
+        } else {
+            return Promise.resolve();
+        }
+    }
+
+    getName() {
+        return remoteConfig.name;
+    }
+
+    getTagNameRegex() {
+        return remoteConfig.tagNameRegex;
+    }
+
+    getPasswordRegex() {
+        return remoteConfig.passwordRegex;
+    }
+
+    getUserNameRegex() {
+        return remoteConfig.userNameRegex;
+    }
+
+    getContactEmail() {
+        return remoteConfig.contactEmail;
+    }
+
+    canSendMails() {
+        return !!remoteConfig.canSendMails;
+    }
+
+    safetyEnabled() {
+        return !!remoteConfig.enableSafety;
+    }
+
     hasPrivilege(lookup) {
         let minViableRank = null;
-        for (let privilege of Object.keys(config.privileges)) {
-            if (!privilege.startsWith(lookup)) {
+        for (let p of Object.keys(remoteConfig.privileges)) {
+            if (!p.startsWith(lookup)) {
                 continue;
             }
-            const rankName = config.privileges[privilege];
-            const rankIndex = this.allRanks.indexOf(rankName);
+            const rankIndex = this.allRanks.indexOf(
+                remoteConfig.privileges[p]);
             if (minViableRank === null || rankIndex < minViableRank) {
                 minViableRank = rankIndex;
             }
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -26,46 +26,51 @@ router.enter(
         next();
     });
 
-// register controller routes
-let controllers = [];
-controllers.push(require('./controllers/home_controller.js'));
-controllers.push(require('./controllers/help_controller.js'));
-controllers.push(require('./controllers/auth_controller.js'));
-controllers.push(require('./controllers/password_reset_controller.js'));
-controllers.push(require('./controllers/comments_controller.js'));
-controllers.push(require('./controllers/snapshots_controller.js'));
-controllers.push(require('./controllers/post_detail_controller.js'));
-controllers.push(require('./controllers/post_main_controller.js'));
-controllers.push(require('./controllers/post_list_controller.js'));
-controllers.push(require('./controllers/post_upload_controller.js'));
-controllers.push(require('./controllers/tag_controller.js'));
-controllers.push(require('./controllers/tag_list_controller.js'));
-controllers.push(require('./controllers/tag_categories_controller.js'));
-controllers.push(require('./controllers/settings_controller.js'));
-controllers.push(require('./controllers/user_controller.js'));
-controllers.push(require('./controllers/user_list_controller.js'));
-controllers.push(require('./controllers/user_registration_controller.js'));
-
-// 404 controller needs to be registered last
-controllers.push(require('./controllers/not_found_controller.js'));
-
-for (let controller of controllers) {
-    controller(router);
-}
-
 const tags = require('./tags.js');
 const api = require('./api.js');
 tags.refreshCategoryColorMap(); // we don't care about errors
-api.loginFromCookies().then(() => {
-        router.start();
-    }, error => {
-        if (window.location.href.indexOf('login') !== -1) {
-            api.forget();
+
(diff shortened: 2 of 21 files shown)


client/js/api.js
```diff
@@ -65,14 +66,49 @@ class Api extends events.EventTarget {
...
-        for (let privilege of Object.keys(config.privileges)) {
-            if (!privilege.startsWith(lookup)) {
+        for (let p of Object.keys(remoteConfig.privileges)) {
+            if (!p.startsWith(lookup)) {
                 continue;
             }
-            const rankName = config.privileges[privilege];
-            const rankIndex = this.allRanks.indexOf(rankName);
+            const rankIndex = this.allRanks.indexOf(
+              remoteConfig.privileges[p]);
```
rr-: Four spaces indentation

client/js/main.js
```diff
@@ -26,46 +26,52 @@ router.enter(
...
+
+    // 404 controller needs to be registered last
+    controllers.push(require('./controllers/not_found_controller.js'));
+
+    for (let controller of controllers) {
+        controller(router);
+    }
+}, error => {
+    console.log('Could not fetch basic configuration from server');
+    // TODO: Better Error Handling
```
rr-: `alert` should do fine

server/szurubooru/api/info_api.py
```diff
@@ -35,11 +35,14 @@ def get_info(
...
         'serverTime': datetime.utcnow(),
         'config': {
+            'name': config.config['name'],
             'userNameRegex': config.config['user_name_regex'],
             'passwordRegex': config.config['password_regex'],
             'tagNameRegex': config.config['tag_name_regex'],
             'tagCategoryNameRegex': config.config['tag_category_name_regex'],
             'defaultUserRank': config.config['default_rank'],
+            'contactEmail': config.config['contactEmail'],
+            'canSendMails': 1 if config.config['smtp']['host'] else 0,
```
rr-: `true` or `false`
`!!config.config['smtp']['host']` should do fine

rr-:
Only cosmetic changes requested. Also please squash to one commit

sgsunder:
Done

rr-:
Thanks

sgsunder:
Well that was silly,

1. `!!str` is a Javascript construct, the python equivalent would be `not not str`, but I've used `bool(str)` cause I think that's more elegant.

2. I also forgot to add the `saftey_enabled` variable from the server fetch.

Both of those changes should be squashed into this PR now.

Right now, the only two variables from `config.yaml` that are needed are the base URL and API URL. I don't see a trivial way around this, so I'm not gonna touch that. Either way, deployment in Docker is much cleaner now!

mserajnik:
> Right now, the only two variables from `config.yaml` that are needed are the base URL and API URL. I don't see a trivial way around this, so I'm not gonna touch that. Either way, deployment in Docker is much cleaner now!

Shouldn't we adjust the `config.yaml.dist` and the `INSTALL.md` accordingly then?

sgsunder:
Yeah, I didn't include any changes to `build.js` because that would break deployment of existing users, but the version of [`build.js`](https://github.com/sgsunder/szurubooru/blob/docker/docker/frontend/build.js) that I'm testing doesn't need either of the `config.yaml` files available to run.

Instead, the base and API URLs, and the build info are set as environment variables:

```
export BASE_URL="http://example.com/"
export API_URL="http://example.com/api/"
export BUILD_INFO=$(git describe --always --dirty --long --tags)
node build.js --debug-client --no-transpile
```

The flags `--debug-client` and `--no-transpile` should usually be left out, however.

The only variable in `config.yaml.dist` that can be removed would be the transpile option, as everything else would either be used by the server, or passed on from the server to the client at runtime.

@rr- I can create another PR for this file if you wish, but it might break the deployment process of some existing users.

rr-:
I don't see the benefit of this approach; why specify base url in two places?

sgsunder:
@rr- 

I was looking through the code for a better solution. Honestly, I'm not sure why the base URL is needed in the first place. 99% of the client-side codebase is already using relative paths for hyperlinks. The only two exceptions are in `client/js/util/markdown.js` line 68:

```javascript
        text = text.replace(
            new RegExp('\\b' + baseUrl + '/post/(\\d+)/?\\b', 'g'), '@$1');
        text = text.replace(
            new RegExp('\\b' + baseUrl + '/tag/([a-zA-Z0-9_-]+?)/?', 'g'),
            '#$1');
        text = text.replace(
            new RegExp('\\b' + baseUrl + '/user/([a-zA-Z0-9_-]+?)/?', 'g'),
            '+$1');
```
which can be replaced with relative URLs:
```javascript
        text = text.replace(
            new RegExp('\\b/post/(\\d+)/?\\b', 'g'), '@$1');
        text = text.replace(
            new RegExp('\\b/tag/([a-zA-Z0-9_-]+?)/?', 'g'), '#$1');
        text = text.replace(
            new RegExp('\\b/user/([a-zA-Z0-9_-]+?)/?', 'g'), '+$1');
```
with no consequences, as well as on `api/password_reset_api.py` line 24, where a similar thing can be done.

The API URL doesn't get used by the server side, and the data URL doesn't get used by the client side scripts, so it would be safe to only provide the client build with only the API URL, and the server with only the data URL, and not have to repeat parameters.

As a side note, right now I'm using 
```yaml
base_url: /
api_url: /api/
data_url: /data/
```
in my deployment, which works fine without any hiccups.

rr-:
It's there because relative links suck.
I'm going to repeat my question and ask, why is this a problem?

sgsunder:
It isn't a problem, or rather it isn't a necessary change to make the code function. If it was I would have submitted it into the original PR.

It's a change that streamlines the build and deployment process in Docker. It may also help the deployment process for other users, but I don't know for sure cause I'm only using Docker. If you don't think that is necessary, that's fine, I'll just stick to my own fork.

The advantage with my method is that the build of the client side can be done completely independent of all configurations, save for one parameter (`api_url`). So if you run the server for static content and the backend API on different containers (which is standard practice for Docker), or completely separate machines, you won't have to duplicate and synchronize configs between them.

The other thing is that relative links are already used elsewhere, so I don't see why this is a problem.
See `models/top_navigation.js` for an example.

rr-:
>See models/top_navigation.js for an example.

…you're right. Now that I actually checked things out, it seems that the whole project uses a mixture of absolute and relative links. It's a mess, really.

I really would have preferred it if the links were absolute, at least when served by the backend. But I guess it's going to be easier

>The advantage with my method is that the build of the client side can be done completely independent of all configurations

I like this approach. But this will need further adjustments in the code. What do you think about following course of action:

* Get rid of `api_url`, `base_url`, `data_url` from the configuration
* Adjust the code base to figure these URLs by hardcoding them
* Consider moving `example.com/api` and `example.com/data` to `api.example.com` and `static.example.com` and require users to add `CNAME` for these subdomains?
* Have backend figure the base url from `HTTP_HOST` user header when serving absolute links to `avatarUrl`, `contentUrl` and `thumbnailUrl`
* Check in `Dockerfile` into source code
* Adjust documentation, replace the whole deployment instructions with instructions how to get the docker image to run

rr-:
Alright, fair point. Let's stick to subdirectories, then.

sgsunder:
Seconded, using separate subdomains would probably break my install because I run it LAN-only, on one machine with a single IP address.

I think the easiest thing to do would be to enforce `api_url = base_url + '/api'`, so `base_url = http[s]://example.com/<opt. prefix>` and `api_url = http[s]://example.com/<opt. prefix>/api`. 

The frontend Docker image that I use is already configured to reverse proxy `/api` URLs to the backend and also serves image content on `/data` by default, which should be fine for small scales.

For larger scales, we can encourage users to use something like HAProxy, which can handle SSL, proxy between a frontend server, API backend server, and load balance between multiple `/data` servers for static content, instead of using multiple subdomains.

Issue #169, opened 2018-06-22
Docker deployment


joshuaavalon:
I am experimenting on docker deployment. Will you accept pull request on that?

joshuaavalon:
I want to confirm is the version compatibility for postgres and elasticsearch. I am trying postgres 10.4 and elasticsearch 6.3.0. It does not seem to have problems.

rr-:
There shouldn't be any problems; I'm running an instance with elasticsearch 6.2.3-1 and postgresql 10.3-1.

joshuaavalon:
@rr- Would you mind to help test it out? https://github.com/joshuaavalon/szurubooru/tree/docker

You need a `config.yaml` at the project folder with  `data_dir: '/data'` and `database: 'postgres://szurubooru:szurubooru@postgres:5432/szurubooru'`.

Then, `docker-compose up -d` will works. Currently there are two Dockerfile for the server. `Dockerfile.server` is based on Debian while `Dockerfile.server.alpine` is based on Alpine.

The Debian version should run fine and fast to build but the image is huge (~1.1GB).
The Alpine version is what worries me. I afraid I may miss some dependency.  The Alpine version is much smaller (~660MB) but it takes a long time to build (about 45 mins on my Digital Ocean VM) because there are no compiled library for Alpine and it needs to compile every libraries in the ``requirements.txt`.

I want you to help testing on the Alpine version.

Also, you can either let the user build the image (docker-compose) or provide a image on DockerHub. What is your opinion on this?

sgsunder:
@joshuaavalon you can look at [my docker deployment](https://github.com/sgsunder/szurubooru/tree/docker), it works similar to yours but incorporates the latest commit. I use `python:3.6-slim` which is much smaller than `python:3.6-jessie`, but doesn't need to build python packages like alpine does because of musl. I also custom build ffmpeg with only the needed features to shrink the image even more. The backend image is about 500MB and the frontend is about 20MB.

Now that the PR I've submitted has been merged, I can probably make another PR for my docker deployment soon, but the smart thing would be to work together with Josh first. I need to add a readme file, too.
The smarter

joshuaavalon:
@sgsunder `python:3.6-slim` is a much better solution. I remove Alpine and Jessie. I skimmed through your docker deployment. I see you extract many configurations to `docker-compose.yml` with extract script but I prefer a simpler approach with at little change to original deployment at possible.

rr-:
[...]

PR #171, opened 2018-07-02
Add .travis.yml

nothink:


rr-:
Wouldn't it be better off to wait for #169 ?

nothink:
Umm... it's certainly right. Please pend it.

rr-:
Docker support has landed

nothink:
OK.
This .travis.yml is based on 2.2 tag.
First of all, is it necessary to match with the current HEAD?

rr-:
Yes, that's encouraged.

nothink:
Umm... Pycodestyle is throwing up some errors, and test_image_hash.py failed.

rr-:
Fixed failing test (regression caused by `HEAD^`).
`pycodestyle` raises known flags:

- `szurubooru/search/configs/post_search_config.py:72:35: E711 comparison to None should be 'if cond is None:'`
  "Fixing" this causes sqlalchemy to fail to produce proper query since it only overloads `__eq__` and not `is`.
- `szurubooru/migrations/env.py:12:1: E402 module level import not at top of file`
  `szurubooru/migrations/env.py:13:1: E402 module level import not at top of file`
  I don't see a clear way to refactor the code that doesn't make it obscure.

nothink:
OK. Should I suppress messages E711, E402, and so on?

rr-:
For time being I'd rather disable `pycodestyle` from the CI build.
For example we also use `pylint` but it too raises some flags that aren't suppressed.

nothink:
All right. Server-side py.test was all passed.

nothink:
Current tests:
 - Backend test for on-premise (3.6/3.7)
 - Check of docker-compose.yml
 - Check of building frontend

Need more?

PR #172, opened 2018-07-06
client: improved build.js, use relative links

sgsunder:
This follows up the discussion on PR #170 

This removes all usage of the `base_url` parameter, by using relative links throughout the codebase. I know that you mentioned that you wanted to potentially return to using absolute links, but given the current state of using a mix of absolute and relative links, this makes things both more consistent and simpler. The relevant changed files are `markdown.js` (inside the `EntityPermalinkWrapper` class) and `password_reset.py`.

The `api_url` parameter is also removed. The `build.js` file instead looks for an environment variable called `API_URL` when run. This is the only configuration that the client code needs at this point. The backend doesn't need this parameter, in fact, it never looked for it in the first place. This means that the `config.yaml` files can now be set later, when the server is run.

To disable the transpiler or enable debugging, you now pass the flags `--no-transpile` or `--debug`. I've also made a git installation optional for the build, you can just set a `BUILD_INFO` variable instead.

I've also included updates to the installation guide, removed some unnecessary require calls to `config.js`, as well as node packages that are no longer needed by `build.js`


sgsunder:
Okay, those changes should be squashed in now

rr-:
What do you think about making `API_URL` default to `/api/` ?

sgsunder:
Yeah, I think that abandoning the API URL parameter altogether would possibly be cleaner. Hardcode the value to `/api/` and ask users to configure `nginx.conf` or a dedicated reverse-proxy instead.

Do you want me to try that on a separate PR?

rr-:
TBH I think it could be done in this PR, especially that essentially `API_URL` is going to be removed just after it was introduced.

sgsunder:
Done

sgsunder:
Added a tiny change (in `client/html/post_readonly_sidebar.tpl` and `client/js/models/post.js`) which makes both relative and absolute data urls work with the IQDB and google image links.

rr-:
Thanks :)

PR #177, opened 2018-07-06
client/build: replace uglify-es, update dependencies

### LINKED ISSUE #174: npm module ' babel-preset-es2015' is deprecated in favor of 'babel-preset-env'
When going through the normal [installation process](https://github.com/rr-/szurubooru/blob/master/INSTALL.md#preparing-environment) and running `npm install`, it gives the following error message (building works fine though):
`npm WARN deprecated babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!`

### LINKED ISSUE #175: npm module 'formidable' v1.0.17 is deprecated
When going through the normal [installation process](https://github.com/rr-/szurubooru/blob/master/INSTALL.md#preparing-environment) and running `npm install`, it gives the following error message (this is the second message overall after the one in issue #174 ):
`npm WARN deprecated formidable@1.0.17: Old versions of Formidable are not compatible with the current Node.js; Upgrade to 1.2.0 or later`

Please forgive me for opening so many issues.

mserajnik:
I've updated all the dependencies and replaced `uglify-es` with [`terser`](https://github.com/fabiosantoscode/terser) (a hard fork of uglify-es, which is apparently no longer actively maintained). The updates got rid of any security issues `npm audit` would report (there were over 300 minor issues before iirc). This closes #174 and closes #175.

Next, I've replaced `babel-preset-es-2015` with `babel-preset-env`. In the near future, we may want to upgrade to [Babel 7](https://babeljs.io/docs/en/next/v7-migration), but that's still in beta so I've left it out for now.

I've also gone ahead and split up the dependencies into regular `dependencies` (which are required for the application to run) and `devDependencies` (which are just required for the build). It doesn't matter right now since npm doesn't really treat `devDependencies` differently in a meaningful way, but it might get relevant in the future when we won't have to bundle up the JS anymore and can instead [use modules directly in the browser](https://developers.google.com/web/fundamentals/primers/modules) reliably. It's also easier to read this way, I guess.

Also, I've just noticed that my home server is down and doesn't come up again (probably a trip switch that went off in my flat, we've had quite a few thunder storms the last days), which means I can't really test this easily right now since I'm on vacation and would need to set up a local installation. So, if someone could test it for me or if one of the Docker experts (@sgsunder) could show me how to spin up a szurubooru container on my MacBook, that would be awesome. Of course, only if it's not too much of a hassle to write down a guide, I don't really know how complicated such a setup is. And unfortunately, I don't have time to read through the Docker docs to learn how to create my own images (never really used Docker before, except for existing images) or instead set up a VM by hand right now – busy vacation. :D

Otherwise, the merge will have to wait until I'm back at home (probably end of next week) and can boot my server up again to test it there.

In any case, the build completes just fine and there weren't any major version changes that would affect the app JS except for `superagent`, but looking through their upgrade guide it doesn't look like we have to adjust anything.

### DIFF
--- a/client/.babelrc
+++ b/client/.babelrc
@@ -1,1 +1,1 @@
-{ "presets": ["es2015"] }
+{ "presets": ["env"] }
--- a/client/build.js
+++ b/client/build.js
@@ -46,11 +46,11 @@ function copyFile(source, target) {
 }
 
 function minifyJs(path) {
-    return require('uglify-es').minify(fs.readFileSync(path, 'utf-8'), {compress: {unused: false}}).code;
+    return require('terser').minify(fs.readFileSync(path, 'utf-8'), {compress: {unused: false}}).code;
 }
 
 function minifyCss(css) {
-    return require('csso').minify(css);
+    return require('csso').minify(css).css;
 }
 
 function minifyHtml(html) {
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -4,63 +4,82 @@
   "lockfileVersion": 1,
   "dependencies": {
     "JSONStream": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
-      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "version": "1.3.3",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.3.tgz",
+      "integrity": "sha512-3Sp6WZZ/lXl+nTDoGpGWHEpTnnC6X5fnkolYZR6nwIfzbxxvA8utPWe1gCt7i0m9uVGsSz2IS8K8mJ7HmlduMg==",
+      "dev": true,
       "requires": {
         "jsonparse": "^1.2.0",
         "through": ">=2.2.7 <3"
       }
     },
     "acorn": {
-      "version": "4.0.13",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
-      "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c="
+      "version": "5.7.1",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.7.1.tgz",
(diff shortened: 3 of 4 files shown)


sgsunder:
For docker, you can just use [my fork](https://github.com/sgsunder/szurubooru) for now, until #169 gets closed. My fork is up to date with this repo now, so you can mirror commits you make on to that as well for testing.

For demoing/development purposes it should be pretty simple.
1. Install Docker for macOS, which should create a small Linux VM for you to run Docker on.
2. Clone my repo, cd to the root of the project
3. You need to follow the instructions in [this link](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-cli-run-prod-mode) for elasticsearch to work. I'm running an older version of Docker for Mac, so I had to do no. 3, but you might need to do no. 2 instead. Not sure.

Run `docker-compose build`. Starting from a clean state, this takes about 30 minutes on my 2014-era MacBook, and this does basically everything in the install guide (fetch node and pip packages, bundles up the client, compiles ffmpeg, etc). If you re-run it after doing changes to the codebase, it should be much faster due to build caching.

For the first run, do `docker-compose up -d db && docker-compose logs -f db` and wait until the database is initialized (should be quick). ctrl+c to stop following the logs. Then do `docker-compose up -d esearch && docker-compose logs -f esearch` and make sure it isn't crashing.

Then run `docker-compose up -d` to start everything else. In later runs you can just skip straight to this part. For my install, I get the VMs IP address with `docker-machine ip`, but I don't know if that will work for you. Then you just open that IP address in a browser. To shut down everything, run `docker-compose down`.

mserajnik:
@sgsunder Thank you. :) `docker-machine ip` didn't work for me (got the error `Error: No machine name(s) specified and no "default" machine exists`, didn't look into it further) but I've noticed that `docker ps` actually listed the correct IP (`0.0.0.0:80`) under `ports`.

I've made a small correction because I had somehow overlooked that the `csso` major version upgrade had some minor usage changes that didn't cause the CSS build to fail but only write `[object Object]` into `app.min.css` – this is now fixed.

Everything else looks good after some testing on my end, uploading, tagging, commenting, searching etc. all worked fine.

@rr-, passing this over to you for testing on your end and approval.

sgsunder:
It's only a deprecation warning, but we should look into closing #178 in this PR as well.

rr-:
I think changing backend dependencies should go into another PR.

rr-:
Everything seems to work well. Thanks :)

PR #180, opened 2018-07-09
Added Docker functionality

### LINKED ISSUE #169: Docker deployment

sgsunder:
Closes #169 

Mostly new files, so there shouldn't be any conflicts with other deployment methods. Only real file change is in `config.py`, which should smartly detect if it's running inside a container and pull some stuff from environment variables, and shouldn't affect anyone running it outside Docker.

The only real point of concern is ffmpeg. I elected to build a slimmed down copy inside the Dockerfile, instead of using a debian package. It takes a lot longer to build, but saves about 150MB in final image size buy cutting out useless stuff like HEVC, subtitle codecs, etc. And if you don't clear the build cache, you usually won't have to rebuild it, even when you rebuild the rest of the image.

I don't really have a strong opinion on this though, so if you think using packages is better, that can be quickly fixed.

### DIFF
--- a/INSTALL-OLD.md
+++ b/INSTALL-OLD.md
@@ -1,0 +1,212 @@
+**This installation guide is deprecated and might be out
+of date! It is recommended that you deploy using
+[Docker](https://github.com/rr-/szurubooru/blob/master/INSTALL.md)
+instead.**
+
+This guide assumes Arch Linux. Although exact instructions for other
+distributions are different, the steps stay roughly the same.
+
+### Installing hard dependencies
+
+```console
+user@host:~$ sudo pacman -S postgresql
+user@host:~$ sudo pacman -S python
+user@host:~$ sudo pacman -S python-pip
+user@host:~$ sudo pacman -S ffmpeg
+user@host:~$ sudo pacman -S npm
+user@host:~$ sudo pacman -S elasticsearch
+user@host:~$ sudo pip install virtualenv
+user@host:~$ python --version
+Python 3.5.1
+```
+
+The reason `ffmpeg` is used over, say, `ImageMagick` or even `PIL` is because of
+Flash and video posts.
+
+
+
+### Setting up a database
+
+First, basic `postgres` configuration:
+
+```console
+user@host:~$ sudo -i -u postgres initdb --locale en_US.UTF-8 -E UTF8 -D /var/lib/postgres/data
+user@host:~$ sudo systemctl start postgresql
+user@host:~$ sudo systemctl enable postgresql
+```
+
+Then creating a database:
+
+```console
+user@host:~$ sudo -i -u postgres createuser --interactive
+Enter name of role to add: szuru
+Shall the new role be a superuser? (y/n) n
+Shall the new role be allowed to create databases? (y/n) n
+Shall the new role be allowed to create more new roles? (y/n) n
+user@host:~$ sudo -i -u postgres createdb szuru
+user@host:~$ sudo -i -u postgres psql -c "ALTER USER szuru PASSWORD 'dog';"
+```
+
+
+
+### Setting up elasticsearch
+
+```console
+user@host:~$ sudo systemctl start elasticsearch
+user@host:~$ sudo systemctl enable elasticsearch
+```
+
+### Preparing environment
+
+Getting `szurubooru`:
+
+```console
+user@host:~$ git clone https://github.com/rr-/szurubooru.git szuru
+user@host:~$ cd szuru
+```
+
+Installing frontend dependencies:
+
+```console
+user@host:szuru$ cd client
+user@host:szuru/client$ npm install
+```
+
+`npm` sandboxes dependencies by default, i.e. installs them to
+`./node_modules`. This is good, because it avoids polluting the system with the
+project's dependencies. To make Python work the same way, we'll use
+`virtualenv`. Installing backend dependencies with `virtualenv` looks like
+this:
+
+```console
+user@host:szuru/client$ cd ../server
+user@host:szuru/server$ virtualenv python_modules # consistent with node_modules
+user@host:szuru/server$ source python_modules/bin/activate # enters the sandbox
+(python_modules) user@host:szuru/server$ pip install -r requirements.txt # installs the dependencies
+```
+
+
+
+### Preparing `szurubooru` for first run
+
+1. Compile the frontend:
+
+    ```console
+    user@host:szuru$ cd client
+    user@host:szuru/client$ node build.js
+    ```
+
+    You can include the flags `--no-transpile` to disable the JavaScript
+    transpiler, which provides compatibility with older browsers, and
+    `--debug` to generate JS source mappings.
+
+2. Configure things:
+
+    ```console
+    user@host:szuru/client$ cd ..
+    user@host:szuru$ mv server/config.yaml.dist .
+    user@host:szuru$ cp config.yaml.dist config.yaml
+    user@host:szuru$ vim config.yaml
+    ```
+
+    Pay extra attention to these fields:
+
+    - data directory,
+    - data URL,
+    - database,
+    - the `smtp` section.
+
+3. Upgrade the database:
+
+    ```console
+    user@host:szuru/client$ cd ../server
+    user@host:szuru/server$ source python_modules/bin/activate
+    (python_modules) user@host:szuru/server$ alembic upgrade head
+    ```
+
+    `alembic` should have been installed during installation of `szurubooru`'s
+    dependencies.
+
+4. Run the tests:
+
+    ```console
+    (python_modules) user@host:szuru/server$ pytest
+    ```
+
+It is recommended to rebuild the frontend after each change to configuration.
+
+
+
+### Wiring `szurubooru` to the web server
+
+`szurubooru` is divided into two parts: public static files, and the API. It
+tries not to impose any networking configurations on the user, so it is the
+user's responsibility to wire these to their web server.
+
+The static files are located in the `client/public/data` directory and are
+meant to be exposed directly to the end users.
+
+The API should be exposed using WSGI server such as `waitress`, `gunicorn` or
+similar. Other configurations might be possible but I didn't pursue them.
+
+API calls are made to the relative URL `/api/`. Your HTTP server should be
+configured to proxy this URL format to the WSGI server. Some users may prefer
+to use a dedicated reverse proxy for this, to incorporate additional features
+such as load balancing and SSL.
+
+Note that the API URL in the virtual host configuration needs to be the same as
+the one in the `config.yaml`, so that client knows how to access the backend!
+
+#### Example
+
+In this example:
+
+- The booru is accessed from `http://example.com/`
+- The API is accessed from `http://example.com/api`
+- The API server listens locally on port 6666, and is proxied by nginx
+- The static files are served from `/srv/www/booru/client/public/data`
+
+**nginx configuration**:
+
+```nginx
+server {
+    listen 80;
+    server_name example.com;
+
+    location ~ ^/api$ {
+        return 302 /api/;
+    }
+    location ~ ^/api/(.*)$ {
+        if ($request_uri ~* "/api/(.*)") { # preserve PATH_INFO as-is
+            proxy_pass http://127.0.0.1:6666/$1;
+        }
+    }
+    location / {
+        root /srv/www/booru/client/public;
+        try_files $uri /index.htm;
+    }
+}
+```
+
+**`config.yaml`**:
+
+```yaml
+data_url: 'http://example.com/data/'
+data_dir: '/srv/www/booru/client/public/data'
+```
+
+To run the server using `waitress`:
+
+```console
+user@host:szuru/server$ source python_modules/bin/activate
+(python_modules) user@host:szuru/server$ pip install waitress
+(python_modules) user@host:szuru/server$ waitress-serve --port 6666 szurubooru.facade:app
+```
+
+or `gunicorn`:
+
+```console
+user@host:szuru/server$ source python_modules/bin/activate
+(python_modules) user@host:szuru/server$ pip install gunicorn
+(python_modules) user@host:szuru/server$ gunicorn szurubooru.facade:app -b 127.0.0.1:6666
+```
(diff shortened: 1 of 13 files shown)


server/szurubooru/config.py
```diff
@@ -14,14 +14,40 @@ def merge(left: Dict, right: Dict) -> Dict:
...
+            'host': os.getenv('POSTGRES_HOST'),
+            'port': os.getenv('POSTGRES_PORT', '5432'),
+            'db': os.getenv('POSTGRES_DB', os.getenv('POSTGRES_USER'))
+        },
+        'elasticsearch': {
+            'host': os.getenv('ESEARCH_HOST'),
+            'port': os.getenv('ESEARCH_PORT', '9200'),
+            'index': os.getenv('ESEARCH_INDEX', 'szurubooru')
+        }
+    }
```
rr-: Blank line needed

server/szurubooru/config.py
```diff
@@ -14,14 +14,40 @@ def merge(left: Dict, right: Dict) -> Dict:
             left[key] = right[key]
     return left
```
rr-: Additional blank line needed

server/szurubooru/config.py
```diff
@@ -1,7 +1,7 @@
 from typing import Dict
 import os
 import yaml
-
+from szurubooru import errors
```
rr-: Additional blank line needed

client/nginx.conf.docker
```diff
@@ -0,0 +1,49 @@
...
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /dev/stdout  main;
+
+    sendfile        on;
```
rr-: The indentation for values in this file is completely arbitrary. I'd either align it in two columns, or make everything use single spaces (like `key value;`). I'm leaning towards the latter solution.

server/.dockerignore
```diff
@@ -0,0 +1,10 @@
+# Don't copy test files over
+szurubooru/tests/*
+setup.cfg
+
+# Other Unneeded Files
```
rr-: I don't think comments in this file add much value

server/szurubooru/config.py
```diff
@@ -14,14 +14,40 @@ def merge(left: Dict, right: Dict) -> Dict:
             left[key] = right[key]
     return left
 
+def docker_config() -> Dict:
+    for key in ['POSTGRES_USER',
+                'POSTGRES_PASSWORD',
+                'POSTGRES_HOST',
+                'ESEARCH_HOST']:
```
rr-:
```diff
-    for key in ['POSTGRES_USER',
-                'POSTGRES_PASSWORD',
-                'POSTGRES_HOST',
-                'ESEARCH_HOST']:
+    for key in [
+            'POSTGRES_USER',
+            'POSTGRES_PASSWORD',
+            'POSTGRES_HOST',
+            'ESEARCH_HOST'
+    ]:
```

server/szurubooru/config.py
```diff
@@ -14,14 +14,40 @@ def merge(left: Dict, right: Dict) -> Dict:
...
+        if not os.getenv(key, False):
+            raise errors.ConfigError('Environment Variable %r not set' % key)
```
rr-:
```diff
-            raise errors.ConfigError('Environment Variable %r not set' % key)
+            raise errors.ConfigError(f'Environment Variable {key} not set')
```

server/szurubooru/config.py
```diff
@@ -14,14 +14,40 @@ def merge(left: Dict, right: Dict) -> Dict:
...
+        'database': 'postgres://%(user)s:%(pass)s@%(host)s:%(port)s/%(db)s' % {
+            'user': os.getenv('POSTGRES_USER'),
+            'pass': os.getenv('POSTGRES_PASSWORD'),
+            'host': os.getenv('POSTGRES_HOST'),
+            'port': os.getenv('POSTGRES_PORT', '5432'),
+            'db': os.getenv('POSTGRES_DB', os.getenv('POSTGRES_USER'))
+        },
+        'elasticsearch': {
+            'host': os.getenv('ESEARCH_HOST'),
+            'port': os.getenv('ESEARCH_PORT', '9200'),
```
rr-: You convert `show_sql` to a bool; shouldn't this get converted to an integer?

server/wait-for-es
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+
```
rr-: Unneeded blank line

server/wait-for-es
```diff
@@ -0,0 +1,32 @@
...
+'''
+Docker helper script. Blocks until the ElasticSearch service is ready.
+'''
+import logging
+import time
+import elasticsearch
+from szurubooru import config, errors
+
```
rr-: Missing blank line

server/wait-for-es
```diff
@@ -0,0 +1,32 @@
...
+    for _ in range(int(TIMEOUT / DELAY)):
+        try:
+            es.cluster.health(wait_for_status='yellow')
+            print('Connected to ElasticSearch!')
+            return
+        except Exception:
+            time.sleep(DELAY)
+            pass
+    raise errors.ThirdPartyError('Error connecting to ElasticSearch')
+
```
rr-: Missing blank line

mserajnik:
> I'd make docker the only deployment process supported, e.g. I'd remove config.yaml.dist in favor of config.yaml.docker (and name the latter just config.yaml.dist); remove BUILD.md etc. etc.
The motivation is that if one wants to build things manually, they can build their own pipeline by examining the docker configuration, which means we have less documentation to support.

I see Docker as a valid _option_ for deployment and it is certainly very useful for testing, but forcing it on people feels wrong to me. E.g., the way my home server is set up, there's no way I could (or rather, would) use Docker for Deployment. I honestly don't see a good reason to do this other than being able to tell people "install with Docker, if it doesn't work for you, read the Docker docs".

What do we lose by keeping the existing build process in addition to adding Docker (and maybe making Docker the default by stating so in the docs)? There's not really a huge level of extra maintenance involved. And the documentation is already there, how often does that really need more than a few additional/corrected lines?

But if you really want to do this, I strongly suggest highlighting this change in some way because this will break any automated build/update process. I'm not sure how many people are actually using szurubooru in production, but with over 100 stars on GitHub I'd assume there are at least a few.

rr-:
The motivation behind making deployment with Docker the only, or at least main supported build process is similar to the reason why there few comments in the source code: rather than talking about how to deploy it through documentation, it'd demonstrate that through working usable code instead. Also it's simpler and requires much fewer steps to go through. It wouldn't break deployment for other people, since they can keep using their workflows, it's just a matter of reworking documentation.

>There's not really a huge level of extra maintenance involved.

There are no tests whether the instructions are true or not, whereas Docker is CI friendly (should we want to use one some day) so it'd be immediately obvious if something was wrong.

sgsunder:
Made the changes. I moved `INSTALL.md` to `INSTALL-OLD.md` and added a deprecation notice at the top, and moved `DOCKER.md` to `INSTALL.md`. Hopefully that's a decent compromise.

sgsunder:
Side note being that Docker images will only work on Linux with x86-64 architecture. So people who deploy on BSD or Windows (I'm sure there's at least a few) or on, say, a Raspberry Pi, won't be able to use it. (If we want ARM support, we would have to make an additional set of Dockerfiles, and I don't think the audience is large enough to make it worth the effort). Sure, they can look through the Dockerfiles as an 'installation guide' of sorts but I wouldn't want to completely break non-Docker installs.

rr-:
>You convert show_sql to a bool; shouldn't this get converted to an integer?

I meant converting `port` to an `int`, not `show_sql` :stuck_out_tongue: 

Anyway, I think these changes are great. I'll try to test these some time this week.

sgsunder:
Added.

It doesn't seem to matter whether the ES port is left as a string or an int. But it'd be best for future debugging headaches for the `docker_config()` function to return the same types as the yaml parser. I think the yaml parser auto-converts number strings to ints, right?

rr-:
>I think the yaml parser auto-converts number strings to ints, right?
Yes

INSTALL-OLD.md
```diff
@@ -0,0 +1,212 @@
+**This installation guide is deprecated and might be out
+of date! It is recommended that you deploy using
+[Docker](https://github.com/rr-/szurubooru/blob/master/DOCKER.md)
```
mserajnik: Since you've renamed `DOCKER.md` to `INSTALL.md`, this link should be adjusted accordingly.

mserajnik:
> It wouldn't break deployment for other people, since they can keep using their workflows, it's just a matter of reworking documentation.

Yup, I misunderstood that part. My comment was based on the (apparently wrong) assumption that building with `config.yaml` wouldn't work "out of the box" anymore and would require people to write their own build process.

Not much else to say, but you have my approval – not that you needed it. :D

> There are no tests whether the instructions are true or not, whereas Docker is CI friendly (should we want to use one some day) so it'd be immediately obvious if something was wrong.

Fair point. On that note, what do you think setting up a simple [Travis CI](https://travis-ci.com/) configuration that runs the backend tests you already have in place and checks if the builds (frontend and backend) succeed? Of course, not in this PR.

rr-:
>On that note, what do you think setting up a simple Travis CI configuration that runs the backend tests you already have in place and checks if the builds (frontend and backend) succeed? Of course, not in this PR.

Actually there's #171 but I told the author to wait until the docker thing gets done.

mserajnik:
> Actually there's #171 but I told the author to wait until the docker thing gets done.

How did I miss that? :|

sgsunder:
Whoops. Fixed now

sgsunder:
I noticed some performance issues on my server, so I scaled down the worker processes in NGINX, and enabled gzip compression. Website feels much snappier now! Also moved the mount directory so that it matches the format of the backend container.

mserajnik:
@sgsunder Thanks, I'll set it up on my machine next week (probably Tuesday or Wednesday) and do some testing, as discussed with @rr- in #181. I've done some Docker "homework" last week (thanks again for your guide on the other PR), so I'm confident I'll get it running. :D

INSTALL.md
```diff
@@ -1,206 +1,65 @@
...
-    `--debug` to generate JS source mappings.
-
-2. Configure things:
+2. Configure the application:
 
     ```console
-    user@host:szuru/client$ cd ..
-    user@host:szuru$ cp config.yaml.dist config.yaml
-    user@host:szuru$ vim config.yaml
+    user@host:szuru$ cp server/config.yaml.docker config.yaml
```
mserajnik: You reference  `server/config.yaml.docker` instead of `server/config.yaml.dist`, but you didn't change the actual filename.

Plus, I'm not sure if copying the file from `server` to the root directory is intended or if you forgot to prepend the target filename with `server/`. If intended, why not just keep the example file in the root directory as well?

INSTALL.md
```diff
@@ -1,206 +1,65 @@
...
+    user@host:szuru$ edit config.yaml
     ```
 
     Pay extra attention to these fields:
 
     - data directory,
     - data URL,
```
mserajnik: The data URL doesn't seem to exist anymore, at least not by default. `data` seems to be the default.

INSTALL.md
```diff
@@ -1,206 +1,65 @@
...
-(python_modules) user@host:szuru/server$ gunicorn szurubooru.facade:app -b 127.0.0.1:6666
-```
+    ```console
+    # To start:
+    user@host:szuru$ docker-compose up -d
+    # To monitor (CTRL+C to exit):
+    user@host:szuru$ docker-compose logs -f
+    # To stop
+    user@host:szuru$ docker-compose down
+    ```
```
mserajnik: All in all, great instructions. The comments in `docker-compose.yml.example` are also really helpful.

The only thing I'd add is the info that the frontend can be reached via `0.0.0.0:port` (the default in `docker-compose.yml` being `0.0.0.0:8080`) as people who haven't worked with Docker before might try `localhost:port` or `127.0.0.1:port` instead.

I see that the frontend is already served via nginx. I'm not that experienced with production deployment with Docker yet, but would you expose the frontend directly on `:80`/`:443` on production and adjust the configuration of the nginx running inside the container to use HTTPS, or would you still expose to `:8080` and then reverse-proxy with an nginx that's running on "bare metal" and has the HTTPS configuration?

I'm mainly curious here, but that might also be something to add (how to best go about deploying for production with HTTPS) – could also be just a link to an external guide.

sgsunder:
Nice catch, made the changes.

The reason that `config.yaml.dist` was moved is because the `server/` directory is used as the root for building the backend, and the builder won't be able to get the file if it's outside that directory. Moving the build context up will make the whole build process a lot messier.

When I run this on my Macbook I actually do use `http://localhost:8080/` to visit the app. Does that not work for you? It should literally just be `http://<IP address/hostname>:<whatever port you set in the compose file>/` and I didn't want to get too hand-holdy (I figured that someone who is able to open up a terminal on their machine can figure out their own local IP).

I didn't write a guide on SSL because I have very little knowledge on that (I personally run this LAN-only). Advanced users who are deploying this public-facing probably would/should make their own compose file anyway.

For reference, [this](https://github.com/sgsunder/docker-compose-nas/blob/master/szurubooru/docker-compose.yml) is how I deploy it on my own NAS.

rr-:
My two cents regarding SSL - I'd delegate this matter to an external nginx instance.
Motivation

- it's not like the performance is going to drop, the bottleneck will always be python (by few orders of magnitude)
- I can host many other sites alongside szurubooru and set them up in a single place

rr-:
Thanks for review @mserajnik , is this ready to be merged?

mserajnik:
@sgsunder You are right, `localhost` does indeed work. `127.0.0.1` doesn't. Weird, I had in my head that I have to use `0.0.0.0`.

Regarding SSL: I guess that's fine, I'm also leaning towards an nginx instance outside of Docker, for the same reasons @rr- has already mentioned. I was just curious because I'm not sure how I'd best handle re-generation of the certificate (in case of Let's Encrypt) etc. if I had my HTTPS-enabled nginx inside a container as well. Mounting the certificate in a volume feels a bit weird to me. Regardless, that's not really part of this PR, so I'm fine with omitting it from the installation guide.

@rr- In my opinion, ready to be merged. 👍

rr-:
Thanks

PR #183, opened 2018-07-21
Clean up required Python packages

### LINKED ISSUE #178: Installation process: python package 'psycopg2' will be renamed to 'psycopg2-binary'
During the installation process while running `alembic upgrade head`, it returns this error first off:
`/home/USERNAME/Path-To-Git-Clone/server/python_modules/lib/python3.5/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.`

I don't know if this even applies to szurubooru, but from a quick look into `requirements.txt`, it shows that indeed this package (`psycopg2`) is used.

sgsunder:
* Packages that are only used in testing or development have been moved to `dev-requirements.txt`
* Closes #178
* Minor rewrite to drop the `scikit-image` package, which saves around 200MB(!) in install size


sgsunder:
Side note,

The RGB to greyscale conversion done by scikit "[is] calibrated for contemporary CRT phosphors" [(src)](http://scikit-image.org/docs/dev/api/skimage.color.html#rgb2grey)

The RGB to greyscale conversion done by Pillow uses the ITU-R 601-2 luma transform [(src)](https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.Image.convert)

From my brief testing, this doesn't seem to make a difference, but we can make Pillow's conversion matrix match scikits exactly if we need to.

rr-:
Merged manually: https://github.com/rr-/szurubooru/commit/9730aa5c050352736327a992f9a428b6bc82cd75

rr-:
I'm worried that Pillow seems to use Floyd-Steinberg dithering and how it might affect the similar image dector's failure rate, but I guess we'll have to see in real life.

PR #187, opened 2018-08-05
client: add basic web app support

### LINKED ISSUE #181: Proposal: add web app manifest
Having a web app manifest would allow users to add szurubooru to the home screen on their mobile devices, making it behave somewhat like a native app in the sense that it could have its own app icon, and opening it would first show a splash screen and then a "fullscreen" browser without status bar (_standalone mode_).

It's basically just a single JSON file (`manifest.json`) that needs to be placed into `public`, plus some stuff we can put inside `<head>` to make it play even nicer (iOS needs some special treatment here).

It would be a quick addition that makes using szurubooru on mobile devices just a bit better. If approved, I can start a PR.

See [here](https://developers.google.com/web/fundamentals/web-app-manifest/) for more information. I would leave out the icon stuff for now, unless you have a high res version of the favicon (let's say 512x512) ready or want to use a different one. The icon would be used as the app icon as well as on the splash screen that would show when opening szurubooru in standalone mode (via the app icon).

If we don't have one, it will just use a screenshot of the application as the app icon, at least that's how it is on iOS. I don't really care because it doesn't affect functionality, but if we had an icon we could use, even better.

mserajnik:
This closes #181.

Chrome picks up the `manifest.json` correctly:

manifest

I've tested it on iOS, [see here](https://www.dropbox.com/s/ck8tqxa7u0msgvr/szurubooru_ios.mov?dl=0) for a video that was taken with the official iOS simulator running an iPhone X. I chose the iPhone X so you can see what happens with the area around the notch – naturally we can't put the app navigation under it, so the best way is just to make it black. Apple doesn't us give [lots of options](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW4) anyway. On other iOS devices without the notch it will of course just fill the whole screen.

As you can see, it integrates quite well. The only real issue I've experienced on iOS is that going back to the home screen will reset the "app" to the initial `/` route. Not sure how that is handled on Android.

Speaking of Android, it should behave pretty similar there (with the splash screen and everything), but should also have a colored browser and app drawer bar (I've used `#24aadd`, which matches the blue used in szurubooru).

If you're curious about why there are so many splash screen images for iOS, I've basically followed [this article](https://medium.com/@applification/progressive-web-app-splash-screens-80340b45d210).

Oh, and let me know if you want those images moved outside the web root. It pretty much just follows the convention of putting the favicon in the web root, but then again, that's from back when it was needed for certain old IEs to pick it up (because the didn't know about `<link rel="icon">`).


rr-:
I'd prefer the images to reside in the `img/` directory. Thanks

mserajnik:
@rr- Done

mserajnik:
Oh, sorry, I misunderstood. I thought you just wanted me to move them from `client/public` to `client/public/img`. They get copied over from `client/app` in the build process.

I'll move them to `client/img` later and adjust the copy.

mserajnik:
Changed. The images are now in `client/img`.

I'm not sure about having the `manifest.json` in `client/app`, but a separate folder makes sense if we think about adding additional web app related files (like service workers) in the future. Plus, this way it doesn't just lay around alongside the build-related files in `client`.

Let me know if you want me to change anything else.

rr-:
Thanks :+1:

PR #194, opened 2018-08-23
Added feature: Specify Base URIs other than `/`

sgsunder:
This allows users to specify a different URI prefix other than `/`: e.g serve the web application on something like `http://example.com/mybooru/` instead of just `http://example.com/`. This allows users to serve mutiple web applications off the same domain name.

This basically leverages on using the HTML `<base>` tag. Docker users set the base href with a `BASE_URL` environment at runtime, while non-docker users can pass it to the `build.js` script. The base href defaults to `/`, so this shouldn't affect the deployment of any existing users. 
The biggest changes were made to `router.js` which reads the base tag from the DOM to figure out the proper routing URIs. The rest of the changes mostly consist of just dropping the leading slash from any URIs to make the base tag work.

For testing: I coded up this feature about a month ago because I needed to run two instances of this on my server. I ran this on both deployments for about a month and haven't noticed any bugs or had any reported to me. However, both instances are run inside a LAN, and only see about 5 users each.

I wouldn't consider myself a frontend/JS dev, so you should inspect these changes carefully to make sure I'm not violating best practices, etc.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -80,9 +80,8 @@ and Docker Compose (version 1.6.0 or greater) already installed.
     Some users may wish to access the service at a different base URI, such
     as `http://example.com/szuru/`, commonly when sharing multiple HTTP
     services on one domain using a reverse proxy. In this case, simply set
-    `BASE_URL="/szuru/"` in the frontend container, and
-    `DATA_URL="/szuru/data/"` in the backend container (unless you are hosting
-    your data on a different domain).
+    `BASE_URL="/szuru/"` in the frontend container (unless you are hosting your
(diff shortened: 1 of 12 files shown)


sgsunder:
I'm not sure what you mean? I [already dropped](https://github.com/rr-/szurubooru/pull/194/files#diff-ddd4eb8decd2223787785f3b3459d12d) the leading slash from the returned string. 

You don't need to do anything more, the browser will automatically interpret the URI as being relative to the `href` from the `<base>` tag.

rr-:
Oh okay, I missed that. Sorry

PR #202, opened 2018-09-22
New flag: check sound

sgsunder:
A lot GIF-like video loops (GFYs or whatever people usually call them) on the internet nowadays come with sound. This allows users to add a sound flag to video posts so that others can know if a post has sound before playing a clip.

First commit enables a sound flag on the backend, shows if it's present in the post view, and allows for setting during upload and post edit.

Second commit allows for flags to be searched, with `flag:sound` and `flag:loop`. The best way to do this was to migrate the flags column in the post table from PickleType/binary to varchar, stored as a comma delimited list. This allows for the column to be queried easily (I'm not sure if it's even possible to query a PickleType in SQLAlchemy, as I can't find any documentation or examples) and also has the side benefit of being human-readable.

Looking through the commit history, I'm not exactly sure why PickleType was used in the first place, so I'm not sure if I'm losing any capabilities with this change.

### DIFF
--- a/client/html/help_search_posts.tpl
+++ b/client/html/help_search_posts.tpl
@@ -71,6 +71,10 @@
             <td>given type of posts. <code>&lt;value&gt;</code> can be either <code>image</code>, <code>animation</code> (or <code>animated</code> or <code>anim</code>), <code>flash</code> (or <code>swf</code>) or <code>video</code> (or <code>webm</code>).</td>
         </tr>
         <tr>
+            <td><code>flag</code></td>
+            <td>having given flag. <code>&lt;value&gt;</code> can be either <code>loop</code> or <code>sound</code>.</td>
+        </tr>
+        <tr>
             <td><code>content-checksum</code></td>
             <td>having given SHA1 checksum</td>
         </tr>
--- a/client/html/post_edit_sidebar.tpl
+++ b/client/html/post_edit_sidebar.tpl
@@ -50,6 +50,11 @@
                     name: 'loop',
                     checked: ctx.post.flags.includes('loop'),
                 }) %>
+                <%= ctx.makeCheckbox({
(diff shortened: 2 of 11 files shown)


rr-:
Can you show a screenshot how the icon looks like with relation to the like/fav count?

sgsunder:
Post edit view:
edit
Post upload view:
upload
Post view:
view

rr-:
Now that I think of it, wouldn't it be better if this flag was set automatically basing on the post content? As far as I remember, the server marks animated gifs that way.

sgsunder:
I thought about about using ffmpeg/ffprobe to detect if there's an an audio track contained in the file. That would work 90% of the time, but the problem is that there are cases where there is an audio track, but it's just zeroed out or contains only really low-level noise. So you would have to do some more detailed processing to determine the intensity of the audio track and set a threshold for which it qualifies as 'having sound.'

I don't necessarily think that's a bad way of going about it, but having the flag be user-set on upload is much simpler and sidesteps the problem.

rr-:
I'll tell you right away that no one is going to bother to set it on our instance.

>Low noise levels

I don't think that's a problem.

How about we remove the checkbox from the upload form, have ffmpeg decide if the file contains audio track after the post is uploaded, and if it makes the wrong decision, the user can still tweak the post afterwards by editing it?

sgsunder:
I've removed the checkbox from the upload form, it's now set automatically on `POST /posts/` by both checking to see if there's an audio track and the volume of the track.

Much easier than anticipated, turns out that ffmpeg has a function called `volumedetect` that can give the average volume of a media file. For posterity: [this blog post](https://medium.com/@SteveBarbera/volume-detection-for-a-audio-stream-dbe727085783) was used as a reference

PR #210, opened 2018-12-13
Use the document base href to generate absolute url

icewind1991:
Otherwise the image link send to IQDB/google images will be invalid


rr-:
Thanks!

Issue #204, opened 2018-10-18
Feature request: Link to source image (Pixiv/DeviantART/Twitter)


Flanamo:
Feature request for the ability to link a source to an image, as in to the original post of the art, like a Pixiv link or a twitter link.

raku-cat:
I noticed when I was messing with the API there's actually a `source` field, and in fact if you update or create a post using the API you can set a source
https://github.com/rr-/szurubooru/blob/master/API.md#creating-post
https://github.com/rr-/szurubooru/blob/master/API.md#updating-post
There's nothing in the webui for this though, I might just be blind though.

sgsunder:
Seems like you're right, this feature is already implemented in the backend. REST calls for this field gets mapped to a `source` column in the `post` table in the database. Just to make sure it's not meant for something else I ran a query `SELECT * FROM post WHERE source IS NOT NULL;`, which returns nothing, so it seems like an unused feature unless you're making external API calls.

Just need to implement it into the frontend then. A simple link called 'Source' can show up in the post view when this field is non-null, next to the reverse image search links. And a simple text field can be used for the post edit view.

rr-:
We used to have this feature, but the users of our instance weren't really using it, so I had removed it back in the days when szurubooru was not so popular. I didn't touch the backend, though.

Flanamo:
Ah so it's just the frontend that's missing it, didn't realize the API even mentioned a source I must've skimmed over it too fast.

raku-cat:
I couldn't find anything in the commit history of where this was removed on the front-end, I have viewing at least working so far on my branch https://github.com/raku-cat/szurubooru/commit/1bfc7170b11d389cfdcf3a72ecf5fc400e6b5bf5
Haven't been able to get actually setting the source working yet though past having the field to do it show up.
![image](https://user-images.githubusercontent.com/1125449/50466261-26911c80-0962-11e9-9143-8c0451c6db3f.png)
![image](https://user-images.githubusercontent.com/1125449/50466266-298c0d00-0962-11e9-860b-aa26b97d6368.png)
![image](https://user-images.githubusercontent.com/1125449/50466270-327cde80-0962-11e9-858a-31a3c2a63578.png)

rr-:
It's in the archived 1.x branch:
https://github.com/rr-/szurubooru/blob/1.x/public_html/templates/post.tpl#L166-L178

PR #215, opened 2018-12-27
Implement #204

raku-cat:
Shows source in posts, allows setting the source as well. Might need input sanitizing.

### DIFF
--- a/client/html/post_edit_sidebar.tpl
+++ b/client/html/post_edit_sidebar.tpl
@@ -93,6 +93,15 @@
             </section>
         <% } %>
 
+        <% if (ctx.canEditPostSource) { %>
+            <section class='post-source'>
+                <%= ctx.makeTextInput({
+                    text: 'Source',
+                    name: 'source',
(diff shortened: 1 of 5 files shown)


rr-:
I'd like to avoid that text-ellipsis, if possible.

sgsunder:
Maybe instead of

> Source: [http://example.com/image?id=foobar](http://example.com/image?id=foobar)

we can just show the domain name

> Source: [example.com](http://example.com/image?id=foobar)

and also we can hide the field entirely instead of showing

> Source: None!

rr-:
Domain name, possibly with `/…`, with a tooltip and a working link. I like that.

raku-cat:
> and also we can hide the field entirely instead of showing
> 
> > Source: None!

My reasoning for doing this was so the page looks consistent, same reason for truncating the URL, so there's not some long URL that makes the page extremely long.

rr-:
Implemented by #223

PR #223, opened 2019-02-08
client: Reimplement post source functionality

### LINKED ISSUE #204: Feature request: Link to source image (Pixiv/DeviantART/Twitter)

sgsunder:
Revisiting the stale PR #215 ( which closes #204 ) with some additional fixes. 

The source field is hidden completely on the view sidebar if unset. The source field will only display the host domain to look more readable. So a source of `http://example.com/image?id=foobar` will display as

> Source: [example.com](http://example.com/image?id=foobar)

on the view sidebar. Additionally a tooltip will show the full URL on hover.

### DIFF
--- a/client/html/post_edit_sidebar.tpl
+++ b/client/html/post_edit_sidebar.tpl
@@ -58,6 +58,16 @@
             </section>
         <% } %>
 
+        <% if (ctx.canEditPostSource) { %>
+            <section class='post-source'>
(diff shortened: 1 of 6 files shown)


rr-:
Thanks.

PR #224, opened 2019-02-08
client/posts: Add some UI icons

sgsunder:
Adds a 'loop' icon in the post view sidebar to indicate if the loop flag is set, similar to how there's an icon to indicate if the sound flag is set.

Also, in the post listing page, the labels 'video', 'animation', and 'flash' have been replaced with [this icon](https://fontawesome.com/v4.7.0/icon/film), which looks cleaner IMO


rr-:
Can I see a screenshot? Thanks

sgsunder:
This is an example video post where both the 'loop' and 'sound' flags are set

Sidebar:
image

Search Listing Thumbnail:
image

rr-:
Looks nice, thanks!

PR #228, opened 2019-02-28
Added feature: Bulk upload via URL list

AkgunFatih:


rr-:
This isn't right. The textarea should be visible only if file dropper is rendered with `allowMultiple: true`, otherwise it should stay a single input.

sgsunder:
Closing because stale, I'm also not sure if this is really a high-demand feature at this point.

PR #233, opened 2019-03-08
Added feature: NGINX reverse proxy documentation, minimal NGINX hardening

alectrocute:


INSTALL.md
```diff
@@ -87,7 +87,7 @@ and Docker Compose (version 1.6.0 or greater) already installed.
     `http://<internal IP or hostname of frontend container>/`. For an NGINX
     reverse proxy, that will appear as:
 
-    ```nginx
+    ```
```
rr-: Why do you keep fiddling with this?

alectrocute: > The commits are still not squashed and they still don't cater to the message style

I'm sorry if I'm a noob, but how does one "squash a commit"?

alectrocute: > Why do you keep fiddling with this?

I guess I accidentally added the word `nginx` in there, so I removed it.

alectrocute: Ah, markdown flavor. Sorry, @rr-. Fixed.

PR #234, opened 2019-03-08
Added feature: NGINX reverse proxy documentation, minimal NGINX hardening

alectrocute:
Okay, @rr- ! This should be better; sorry about that. :)


rr-:
Merged manually.

PR #236, opened 2019-03-11
Feature Improvement : Automatic child relations

colin969:
Checks to add the children of newly added relations.

Example:
Posts 1 and 2 are related.
Post 3 creates a relation with Post 1, a relation with Post 2 is added automatically.

### DIFF
--- a/server/szurubooru/func/posts.py
+++ b/server/szurubooru/func/posts.py
@@ -581,8 +581,18 @@ def update_post_relations(post: model.Post, new_post_ids: List[int]) -> None:
     if post.post_id in new_post_ids:
         raise InvalidPostRelationError('Post cannot relate to itself.')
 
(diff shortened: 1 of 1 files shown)


rr-:
This simplifies the relations feature into something like post pools.
I think it's wise to keep the relation graph directed, for example for stuff such as comics / manga that's of sequential nature.

colin969:
How about a parameter that will include child relations inside the response? (rather than changing the relations of the post itself)

It'd allow for choosing of viewing it as a post pool (which is basically all my content) or ordered content like manga, without changing the relations themselves.

rr-:
Yes, sounds like better choice.

sgsunder:
Stale PR

fsnowdin:
bumping as this would be incredibly useful

PR #242, opened 2019-04-07
Add source searching

sgsunder:
Allows searching for posts by source URL.

You can use either something like `source:twitter` to find all all twitter related posts or `source:twitter.com/some-specific-url` for specific URL queries.


rr-:
Thanks

Issue #243, opened 2019-04-08
Enhancement: Left and Right Swipe Gestures On Post Images


ReAnzu:
Using Szurubooru is pleasant on the desktop, on mobile it would be more natural to be able to move between posts using a swipe gesture.

rr-:
I believe this can be now closed after your enhancement was merged :)

ReAnzu:
Thanks!

PR #244, opened 2019-04-08
client/post: Add swipe left and swipe right gestures to post content

ReAnzu:
#243
* Add util/touch.js to abstract the swipe implementation, with support for left, right, up and down baked in.
* Update views/post_main_view to instantiate a new Touch instance to handle left and right swipes on the post image.

### DIFF
--- a/client/js/util/touch.js
+++ b/client/js/util/touch.js
@@ -1,0 +1,85 @@
+const direction = {
+    NONE: null,
+    LEFT: 'left',
+    RIGHT: 'right',
+    DOWN: 'down',
+    UP: 'up'
+};
+
+function handleTouchStart(handler, evt) {
+    const touchEvent = evt.touches[0];
+    handler._xStart = touchEvent.clientX;
+    handler._yStart = touchEvent.clientY;
+}
+
+function handleTouchMove(handler, evt) {
+    if (!handler._xStart || !handler._yStart) {
+        return;
+    }
+
+    const xDirection = handler._xStart - evt.touches[0].clientX;
+    const yDirection = handler._yStart - evt.touches[0].clientY;
+
+    if (Math.abs(xDirection) > Math.abs(yDirection)) {
+        if (xDirection > 0) {
+            handler._direction = direction.LEFT;
+        } else {
+            handler._direction = direction.RIGHT;
+        }
+    } else {
+        if (yDirection > 0) {
+            handler._direction = direction.DOWN;
+        } else {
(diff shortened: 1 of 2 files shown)


rr-:
How does it play with drawing post notes in edit mode?

ReAnzu:
Here's the breakdown across 4 devices. Desktop (Chrome and Firefox), Laptop with touchscreen (Firefox), iPhone (Mobile Safari), iPad (Mobile Safari)

Desktop: Behaves normally, no touch input registered.

Laptop: I can create points with single taps, Any swipe after first tap to set the initial point warns about leaving a page with unsaved form data. I have to hit enter to finish the note location. After hitting enter I can't move the note around (This also happens on the latest build from master, touch events don't move the note around either, I have to use a mouse to move it). Interacting with notes using single touches works normally, no dragging using touches to select (This is identical to the current master).

iPhone: Creating points with single taps works just fine. It's easier to accidentally swipe and get the warning about leaving the page. Tapping on or off a note behaves the same as the laptop.

iPad: Same as the iPhone, but not as easy to accidentally swipe.

It's annoying enough that I updated the PR with a change that disables swiping in edit mode.

rr-:
Thanks :)

client/js/util/touch.js
```diff
@@ -0,0 +1,85 @@
...
+        if (xDirection > 0) {
+            handler._direction = direction.LEFT;
+        } else {
+            handler._direction = direction.RIGHT;
+        }
+    } else {
+        if (yDirection > 0) {
+            handler._direction = direction.DOWN;
+        } else {
+            handler._direction = direction.UP;
```
Hunternif: When testing this functionality, I found that up and down are actually reversed here.

ReAnzu:
I haven’t really noticed it being too eager, but a threshold might be a nice feature. I expect it might have to be dynamic to the resizing to the display elements it’s bound to, which is why I don’t think I tackled that feature immediately. But I’ll play around with it.

Interestingly I never tested up and down as I didn’t have a good idea of what to use those gestures for, but wanted the touch controller to be generic if they were a good idea somewhere else. But logically the y axis is correct. A positive y movement (starting y is less than the ending y, resulting in starty - endy < 0.)

I’ll throw some logging in while I test ideas for a swipe deadzone. I don’t know if any devices/browsers that report the y axis inverted for touches but that’s the type of thing I could see being the case.

What were you thinking of using the up/down gestures for within the app?

PR #245, opened 2019-04-08
Added feature / changed bevavior: Button to randomize search results

Hunternif:
I'm using szurubooru as a sort of a personal image gallery. When looking at images, I like to shuffle them around. Currently in szurubooru the option `sort:random` shuffles only once and then returnes a cached result (from what I understood, it's actually cached both on the client and on the server). I suspect this was done for a good reason, but it doesn't work for me, so I changed it:
1. Changed client and server so that the search queries with `sort:random` are never cached, both on client and server.
2. Added a little button to randomize the current query. Here's what it looks like:
image

I'm not very familiar with this project yet, nor with its stack of technologies, especially JS. Constructive criticism is super welcome.

### DIFF
--- a/client/html/posts_header.tpl
+++ b/client/html/posts_header.tpl
@@ -4,8 +4,9 @@
         %><wbr/><%
         %><input class='mousetrap' type='submit' value='Search'/><%
         %><wbr/><%
-        %><button id="random-button" style="padding:0.2em; margin-right:0.25em; margin-bottom:0.25em;"><i class="fa fa-random"></button><%
-        %><wbr/><%
+        %><button id="random-button" style="padding:0.2em; margin-right:0.25em; margin-bottom:0.25em;"><%
+            %><i class="fa fa-random" style="vertical-align: middle"><%
+                %></button><%
         %><% if (ctx.enableSafety) { %><%
             %><input data-safety=safe type='button' class='mousetrap safety safety-safe <%- ctx.settings.listPosts.safe ? '' : 'disabled' %>'/><%
             %><input data-safety=sketchy type='button' class='mousetrap safety safety-sketchy <%- ctx.settings.listPosts.sketchy ? '' : 'disabled' %>'/><%
--- a/client/html/posts_header.tpl
+++ b/client/html/posts_header.tpl
@@ -4,6 +4,8 @@
         %><wbr/><%
         %><input class='mousetrap' type='submit' value='Search'/><%
         %><wbr/><%
+        %><button id="random-button" style="padding:0.2em; margin-right:0.25em; margin-bottom:0.25em;"><i class="fa fa-random"></button><%
+        %><wbr/><%
         %><% if (ctx.enableSafety) { %><%
             %><input data-safety=safe type='button' class='mousetrap safety safety-safe <%- ctx.settings.listPosts.safe ? '' : 'disabled' %>'/><%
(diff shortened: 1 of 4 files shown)


rr-:
Please squash the commits and amend the commit messages so that they follow the established style.

client/js/views/posts_header_view.js
```diff
@@ -242,6 +247,13 @@ class PostsHeaderView extends events.EventTarget {
         e.preventDefault();
         this._navigate();
     }
+    _evtRandomButtonClick(e) {
```
rr-: One blank line between functions

client/js/views/posts_header_view.js
```diff
@@ -242,6 +247,13 @@ class PostsHeaderView extends events.EventTarget {
...
+        e.preventDefault();
+        if (!this._queryInputNode.value.includes('sort:random')) {
+            this._queryInputNode.value += ' sort:random';
```
rr-: Please trim the leading space (the case when the search query was empty).

server/szurubooru/search/executor.py
```diff
@@ -94,7 +94,7 @@ def execute(
                 disable_eager_loads = True
 
         key = (id(self.config), hash(search_query), offset, limit)
-        if cache.has(key):
+        if not disable_eager_loads and cache.has(key):
```
rr-: This change is good but the code is not very intuitive. Please add a note for me to refactor this in form of `  # TODO: refactor this` before the condition check.

client/js/models/post_list.js
```diff
@@ -17,13 +17,18 @@ class PostList extends AbstractList {
     }
 
     static search(text, offset, limit, fields) {
+        //For queries with random sorting, bypass cache by appending random number
```
rr-: Space after `//`, lowercase comments

client/html/posts_header.tpl
```diff
@@ -4,6 +4,9 @@
         %><wbr/><%
         %><input class='mousetrap' type='submit' value='Search'/><%
         %><wbr/><%
+        %><button id="random-button" style="padding:0.2em; margin-right:0.25em; margin-bottom:0.25em;"><%
```
rr-: No inline styling please

server/szurubooru/search/executor.py
Hunternif: But there's still the issue that this breaks bulk editors. Maybe I should change my approach, and be more explicit about saving post order in client session?
Or I could modify the client again, to not bypass cache when doing bulk edits.

Hunternif: I just realized that this also breaks pagination. The request for the next page is randomized, so it can contain some posts from the previous pages.

rr-: Yes, I talked about it in some other issues. TBH I don't care about this anymore.

Hunternif: But surely you wouldn't want to merge code that introduces blatant bugs like this?

Anyway, I think I figured it out: I'll store `cacheNumber` in context, and only update it when the user clicks the "randomize" button. That should solve all the above issues. I will overwrite this PR.

rr-: Yes, by merging I meant after the bugfixes, of course. I like the idea of keeping the cache number.

Hunternif: Wait, it looks like pagination with `sort:random` is already broken at `master`: some posts appear again on subsequent pages.

rr-: Yes, the post do reappear due to the way the `ORDER BY` clause is formed, but the pagination remains consistent due to backend caching.

PR #247, opened 2019-04-10
Added feature: set tag categories for new tags

Hunternif:
When typing new tags in the post sidebar, I'd like to be able to set their category. I updated TagInputControl to parse the format `category:name`
I didn't change the API contract, so I keep track of the newly added tags with categories, and save them after the post has been saved.
Tags with unknown categories are assigned the default category, and an error message will show up.


rr-:
Sounds like a very obscure feature.

client/js/controllers/post_main_controller.js
```diff
@@ -153,6 +154,25 @@ class PostMainController extends BasePostController {
...
+                let tagPromises = [];
+                for (let newTag of e.detail.newTags) {
+                    // load the tag that was created during updating the post
+                    tagPromises.push(
+                        Tag.get(newTag.names[0]).then(tag => {
+                                tag.category = newTag.category;
+                                return tag.save();
+                            }
+                        ).then(() => {
+                            e.detail.newTags.removeByName(newTag.names[0]);
```
Hunternif: I'm a bit concerned if it's ok to save`e` in closure.
And also if it's ok to create promises to remove from array, while I'm iterating the array.

Hunternif:
This may well be an obscure feature.
I use many categories, and I very often find myself navigating between the post and the Tags page, to change category for the tag I've just created or am about to create.

client/js/controls/post_edit_sidebar_control.js
```diff
@@ -338,6 +338,8 @@ class PostEditSidebarControl extends events.EventTarget {
                     misc.splitByWhitespace(this._tagInputNode.value) :
                     undefined,
 
+                newTags: this._tagControl.newTags,
```
Hunternif: TODO: add a null check here

client/js/controllers/post_main_controller.js
```diff
@@ -153,6 +154,25 @@ class PostMainController extends BasePostController {
         post.save()
             .then(() => {
                 this._view.sidebarControl.showSuccess('Post saved.');
+                // now save the new tags with categories
+                let tagPromises = [];
+                for (let newTag of e.detail.newTags) {
```
Hunternif: TODO: add a null check here

sgsunder:
Can you add help documentation for this somewhere in the HTML help templates?

Hunternif:
Sure, will do. Also I need to update the bulk tag editor to have the same functionality.

client/js/controls/tag_input_control.js
```diff
@@ -139,27 +141,44 @@ class TagInputControl extends events.EventTarget {
     }
 
     addTagByText(text, source) {
+        // try to find category for new tags in the format "category:name"
+        text = text.replace(/:\s+/, ':');
         for (let tagName of text.split(/\s+/).filter(word => word).reverse()) {
-            this.addTagByName(tagName, source);
+            let nameAndCat = tagName.split(':');
+            if (nameAndCat.length > 1) {
+                // "cat:my:tag" should parse to category "cat" and tag "my:tag"
```
Hunternif: Probably it's better to require escaped colons in tag name, i.e. `cat:my\:tag`

OstojaOfficial:
Are there any updates on this?

Issue #248, opened 2019-04-11
Unknown error (delete_by_query() got an unexpected keyword argument ‘doc_type’).


warsix4:
I am unable to upload any type of file to my szurubooru installation. The error I receive is as in the title: 

`Unknown error (delete_by_query() got an unexpected keyword argument ‘doc_type’).`

Any suggestions? Thanks!

warsix4:
After further testing, this only happens with `.jpg` and `.png` files.

Hunternif:
I encountered the same error when running unit test `test_image_hash.py`

Hunternif:
I'm guessing this is because of an incompatible version of the `elasticsearch` library. @warsix4 did you install szurubooru with Docker?

warsix4:
Yes I did. Is there any specific logs or info I can provide?
___


Cody Warmbo
warmbo.com

Hunternif:
I'm speculating as I'm not super familiar with these technologies, but perhaps you could try and install again on a clean system and share the output fom these steps:
```
docker-compose pull
docker-compose build --pull
```

warsix4:
I'd like to resort from recreating a VM but will do so in a few hours I guess if I can't find any solutions or at least causes.

warsix4:
I spun up a new Ubuntu Server 18.04 instance and installed docker/docker-compose. During the 

`docker-compose build --pull` command I get the error:   
'elasticsearch-dsl 6.3.1 has requirement elasticsearch<7.0.0,>=6.0.0, but you'll have elasticsearch 7.0.0 which is incompatible.'

This time after installation I also am unable to actually start the sql container... /var/local/szurubooru does not seem to exist.

warsix4:
Now after ANOTHER attempt I am facing this issue

```
urubooru/sql': mkdir /var/local/szurubooru: read-only file system"

ERROR: for sql  Cannot start service sql: b"error while creating mount source path '/var/local/szurubooru/sql': mkdir /var/local/szurubooru: read-only file system"
```
I have chmod'd the directory, and run everything as sudo, but no go. :( 

I really wish the documentation was more specific or suggested a distro to use.

warsix4:
I have spun up a Debian9 VM instead of Ubuntu this time and now face the exact same error as in the post title again.

Hunternif:
[...]

PR #251, opened 2019-04-16
server/build: Ensure elasticsearch library is a compatible version

Hunternif:
It seems that Elasticsearch version 7.0.0 introduced a breaking change, which caused #248 
Updated server requirements.txt


sgsunder:
Is there a reason why we can't do this:
```
elasticsearch>=5.0.0,<7.0.0
elasticsearch-dsl>=5.0.0,<7.0.0
```

Hunternif:
Excellent idea, let's do that.

PR #252, opened 2019-04-16
server: Fix snapshots and unit tests related to post flags

Hunternif:
TODO: scratch this and change flags to be an array.

Test still failing: `test_get_user_tokens`


Hunternif:
But what about https://github.com/rr-/szurubooru/commit/3879c2ec20b1fa09408bbe7ba3bb93a35fab5796

rr-:
This wasn't right, either.

Hunternif:
Stored as a string in the database, but as an array in the application? Or do you think flags should be a table?

Hunternif:
Don't "fix" these tests, should fix the implementation instead.

PR #261, opened 2019-05-22
Add setting to display underscores as spaces in tags

### LINKED ISSUE #258: [Suggestion] Allow spaces in tags
I think the title is pretty self-explanatory. It'd be very welcome to be able to use spaces in tags.

Underscores are the work-around, I'm well aware, but that's exactly what they are. A work-around.

neobooru:
Fixes #258

Honestly I'm not a huge fan of how this turned out, especially with the 'global' tagUnderscoresAsSpaces variable in `util/misc.js`, but this is the best thing I could come up with to avoid calling `settings.get()` multiple times (which is a somewhat expensive call).
I also didn't fully comment my code as that wasn't done anywhere else in the project so I couldn't copy a documentation style.

This setting is disabled by default, so nothing changes for existing users.


sgsunder:
Thanks!

I think I can author a method to address the expensive `settings.get()` call via caching, too.

sgsunder:
Settings caching done in e0fc790

Issue #256, opened 2019-04-29
Bug: Automatic password recovery emails are missing the base URI


TiredSysOp:
Will also bring up #193 as well, having a option to enter a domain or subdomain in the config is a nice workaround.

![bug](https://user-images.githubusercontent.com/31832099/56880020-0463f400-6a53-11e9-96a2-d2e379ed9e43.png)

The booru has it's name set to a name that will resolve to itself if entered in an address bar, for the life of me I can't get the base uri to show in the link in the password recovery emails

Currenly using latest docker version and it sits behind nginx, using remote smtp server

Cheers

neobooru:
I solved this issue for me as seen in the commit linked above (https://github.com/neobooru/szurubooru/commit/47760de37c5a57ed4214bf22fe88a5039df8f746), however this solution changes the config.yml in some 'breaking' ways, which is why I haven't opened a pull request yet. Any thoughts about this solution, or possible alternatives?

This would also solve #193

sgsunder:
@neobooru I think that your solution is a good idea, I would open up a PR!

sgsunder:
Resolved in #265 and 80d272d

PR #265, opened 2019-05-29
Add 'domain' and 'smtp from' config entries

### LINKED ISSUE #193: Bug: Configure reset password mail sender address
https://github.com/rr-/szurubooru/blob/1fe22a4d0aa6f7e374394afc7777361e48f21823/server/szurubooru/api/password_reset_api.py#L32

Using name as the sender is not really suitable because `name` is usually not a valid domain name. Also, if we can add a config to allow user to set it to match the DNS records, these will reduce the chance of the mail going to spam.

neobooru:
Fixes #193 and #256

This however requires users to manually set the domain and no-reply address in the config.yaml.
The `domain` field currently is optional, but it would probably be better to make it required and not fall back to HTTP_ORIGIN and HTTP_REFERER, which might be inaccurate or not set (especially behind reverse proxies and the like). 

The `smtp:from` field is required, the server will crash without it set.

This also should break any docker configs that currently exist (though I haven't looked at those).

### DIFF
--- a/server/config.yaml.dist
+++ b/server/config.yaml.dist
@@ -3,6 +3,8 @@
 
 # shown in the website title and on the front page
 name: szurubooru
+# full url to the homepage of this szurubooru site, with no trailing slash
+domain: http://example.com
 # user agent name used to download files from the web on behalf of the api users
 user_agent:
 # used to salt the users' password hashes
@@ -32,7 +34,9 @@ smtp:
     port: # example: 25
     user: # example: bot
     pass: # example: groovy123
-    # host can be left empty, in which case it is recommended to fill contactEmail.
+    from: # example: noreply@example.com
+    # if host is left empty the password reset feature will be disabled, in which case it is
+    # recommended to fill contactEmail so that users know who to contact when they want to reset their password
 
 contact_email: # example: bob@example.com. Meant for manual password reset procedures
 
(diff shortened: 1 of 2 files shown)


server/config.yaml.dist
```diff
@@ -3,6 +3,8 @@
 
 # shown in the website title and on the front page
 name: szurubooru
+# full url to the homepage of this szurubooru site, with no trailing slash
+domain: http://example.com
```
sgsunder:
```suggestion
domain: # example: http://example.com
```

Probably better to leave it blank by default

sgsunder:
Sorry it took me so long to get to this.

I'm not sure what you mean by:
> The `smtp:from` field is required

I'm guessing you only mean it's required if you have email-based password resets enabled (i.e. `smtp:host` is set), correct? All the SMTP fields should remain optional if the user only opts for manual password resets.

For the users who do use email-based resets, perhaps we can generate a more friendly and descriptive exception letting users know that `smtp:from` is a new, added parameter that should be set if they haven't already.

neobooru:
Seems fair. Is there already a certain way to throw exceptions in this codebase, or should I just do a descriptive `raise Exception(msg)` when `smtp:from` isn't set? And should that throw on startup or when the user actually tries to send a mail (on startup would probably be better).

sgsunder:
You should probably add the check to `validate_config()` in `facade.py` (which is the application entrypoint, so it would be at startup). Seems like there's already a `errors.ConfigError` defined which you can throw.

If you want to display warnings about stuff like `domain` and such, you could probably call `logging.warning()` within the same function.

sgsunder:
Merged manually with 80d272d , thanks!

Issue #268, opened 2019-07-22
Bug: When logging out and refreshing the page it throws an "Invalid token." error


neobooru:
![image](https://user-images.githubusercontent.com/50623835/61667634-cafdf600-acda-11e9-8066-dbeb4be83eb4.png)

This is probably caused by the auth cookies. Deleting those two fixes all issues. I might just fix the issue myself, but putting it on the board for consistency.

![image](https://user-images.githubusercontent.com/50623835/61667796-42cc2080-acdb-11e9-92b3-2dc0390c90cf.png)

PR #269, opened 2019-07-22
client/tag_categories: load tag_categories after (attempted) login

### LINKED ISSUE #262: Bug: Client does not send auth tokens when calling GET /api/tag-categories
Hi there,

I'm running an updated szurubooru instance and the tags aren't displayed in the color linked to their category, but in the default blue color. 
However, if I go to the tag categories and save the changes again (without changing anything), tags will have their proper color, until I reload the page. Navigating through the booru won't break the colors until the page is completely reloaded again.

neobooru:
Fixes #262 (as far as I could test at least)

What was wrong:
The client called `refreshCategoryColorMap()` before logging in.

The server received the request before `ctx.user` was set, causing `verify_privilege(ctx.user, 'tag_categories:create')` to be called with the anonymous user, and understandably threw an error when the privilege required something higher than 'anonymous'.

Now it calls `refreshCategoryColorMap()` after logging in, after which it calls `verify_privilege()` on the server with the correct `ctx.user`.

### DIFF
--- a/client/js/main.js
+++ b/client/js/main.js
@@ -28,7 +28,6 @@ router.enter(
 
 const tags = require('./tags.js');
 const api = require('./api.js');
-tags.refreshCategoryColorMap(); // we don't care about errors
 
 api.fetchConfig().then(() => {
     // register controller routes
@@ -61,6 +60,7 @@ api.fetchConfig().then(() => {
     window.alert('Could not fetch basic configuration from server');
 }).then(() => {
     api.loginFromCookies().then(() => {
+            tags.refreshCategoryColorMap();
             router.start();
         }, error => {
             if (window.location.href.indexOf('login') !== -1) {


sgsunder:
I'm not super familiar with the logic of the login function calls of the client.

Does it still call `refreshCategoryColorMap()` if the user logs in with a password instead of an auth token?

neobooru:
Login method doesn't matter as far as I am aware (and when clicking remember password it always uses a automatically generated token).

That said I did find a bug with this. When you login and the tags require an authenticated user it'll still error with "Insufficient privileges to do this.". A simple F5/refresh fixes this, but it's still a bug. (This might have something to do with #268)

sgsunder:
Merged manually in 8f0835f , thanks!

When you get a chance, can you open up an issue regarding the "Insufficient privileges to do this." error? We'll track and fix that issue on a seperate ticket.

PR #270, opened 2019-07-22
client/upload: automatically set source when uploading from url

### LINKED ISSUE #230: Feature request: Automatically set source to image URL
Related: #204 and #215 

With the source feature being added/reimplemented into the frontend, it would be nice to automatically set the source for a post to the upload URL if the post in question is uploaded through an URL.

neobooru:
Fixes #230


sgsunder:
Thanks!

PR #271, opened 2019-07-24
client/auth: call tags.refreshCategoryColorMap() after login

neobooru:
When the tag category list permission is not anonymous the category colors fail to load if you are not logged in, and because the page doesn't reload (SPA) the tag colors are still broken after logging in. Manually calling refreshCategoryColorMap after logging in solves this issue.

Should fix last issue discussed in #269


sgsunder:
Thanks!

PR #279, opened 2019-09-25
server/info: report correct size when filesystem is missing files

neobooru:
In case the data directory has missing files (e.g when accidentally deleted from disk - idk who would be this dumb) szurubooru errors as soon as you open the web page (because it calls /api/info which tries to calculate the total_size). This fixes that by just ignoring deleted files.

### DIFF
--- a/server/szurubooru/api/info_api.py
+++ b/server/szurubooru/api/info_api.py
@@ -22,7 +22,7 @@ def _get_disk_usage() -> int:
             file_path = os.path.join(dir_path, file_name)
             try:
                 total_size += os.path.getsize(file_path)
-            except:
+            except FileNotFoundError:
                 pass
     _cache_time = now
     _cache_result = total_size
--- a/server/szurubooru/api/info_api.py
+++ b/server/szurubooru/api/info_api.py
@@ -20,7 +20,10 @@ def _get_disk_usage() -> int:
     for dir_path, _, file_names in os.walk(config.config['data_dir']):
         for file_name in file_names:
             file_path = os.path.join(dir_path, file_name)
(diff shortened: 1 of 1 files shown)


server/szurubooru/api/info_api.py
```diff
@@ -20,7 +20,10 @@ def _get_disk_usage() -> int:
     for dir_path, _, file_names in os.walk(config.config['data_dir']):
         for file_name in file_names:
             file_path = os.path.join(dir_path, file_name)
-            total_size += os.path.getsize(file_path)
+            try:
+                total_size += os.path.getsize(file_path)
+            except:
```
sgsunder: Probably should only catch `FileNotFoundError` so that unrelated errors will still raise an exception

server/szurubooru/api/info_api.py
```diff
@@ -20,7 +20,10 @@ def _get_disk_usage() -> int:
     for dir_path, _, file_names in os.walk(config.config['data_dir']):
```
sgsunder: I'm guessing this is the root cause of the problem:

A race condition occurs between when `os.walk(config.config['data_dir'])` is called and when all the file sizes are summed up at the end of the loop. If someone deletes a post during this period, then that will cause the missing file exception.

Not an issue with this PR specifically, but I'm wondering if there's a smarter way to sidestep the problem entirely...

neobooru:
I vaguely recall it crashing even without that race condition, but looking back now I can't think of a single scenario that would make that possible (maybe my filesystem was corrupt?). Anyway, this should solve that (rather hard to get) race condition.

sgsunder:
Merged manually in d2a4e506, thanks!

PR #280, opened 2019-09-29
client/login: always store login cookie as 'auth'

### LINKED ISSUE #268: Bug: When logging out and refreshing the page it throws an "Invalid token." error

neobooru:
I don't know why the decision was made to use a calculated name for the cookie, and I haven't fully looked at the complete login flow. But this seems to at least fix #268. It might break something (as I don't fully understand the login flow) but everything seems to work fine.

### DIFF
--- a/client/js/api.js
+++ b/client/js/api.js
@@ -9,18 +9,6 @@ const uri = require('./util/uri.js');
 let fileTokens = {};
 let remoteConfig = null;
 
-function getCookieName() {
-    const bases = document.getElementsByTagName('base');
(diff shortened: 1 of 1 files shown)


rr-:
Yes, putting domain name in the cookie name seems very weird.

sgsunder:
Merged manually in c9eae00c, thanks!

PR #283, opened 2019-10-05
Add support for webp images

neobooru:
Couldn't actually run the pytest but it works on my test servers (both nginx and apache, not that this really matters).


neobooru:
Hmm, guess that didn't work, though I can't actually see the dockercloud log.

sgsunder:
Stack trace from pytest:

```python
================================================================== FAILURES ===================================================================
__________________________________________________ test_get_mime_type[webp.webp-image/webp] ___________________________________________________
szurubooru/tests/func/test_mime.py:16: in test_get_mime_type
    assert mime.get_mime_type(read_asset(input_path)) == expected_mime_type
szurubooru/tests/conftest.py:255: in get
    with open(path, 'rb') as handle:
E   FileNotFoundError: [Errno 2] No such file or directory: '/opt/app/szurubooru/tests/assets/webp.webp'
```

So I think you would need to add a sample webp file at `./server/szurubooru/tests/assets/webp.webp`

I'm also assuming that you checked the reverse search functionality, such that:
1) webp files get hashed properly
2) Uploading a webp file that's similar to an existing image is detected
3) Uploading a non-webp file that's similar to an existing webp file is detected

Unfortunately this part can't be tested via pytest so it would need to be checked manually.

_________

For an easy way to run pytest yourself, you can add the following to `.git/config`:
```
[alias]
	build-client = !docker build -t "szurubooru-client:$(git rev-parse --short HEAD)" "$(git rev-parse --show-toplevel)/client/"
	build-server = !docker build -t "szurubooru-server:$(git rev-parse --short HEAD)" "$(git rev-parse --show-toplevel)/server/"
	test-server = !IMAGE_NAME="szurubooru-server:$(git rev-parse --short HEAD)" "$(git rev-parse --show-toplevel)/server/hooks/test"
```

So now you can simply run `git build-server` followed by `git test-server` to test your code. I might add a proper script (probably as a Makefile) sometime soon to do this better.

neobooru:
1 - If you mean the checksum in the database, then yes  
2/3 - Yes, however when the image is exactly the same but a different file type it'll say 100% match instead of displaying the "post already exists" error message (but this is probably always the case, because the checksum is different due to the file header etc).

Will fix the tests tomorrow.

neobooru:
I think this should do it?

sgsunder:
Merged in 73c53fa

Not really sure why Docker doesn't re-test when new commits are added...

PR #286, opened 2019-10-14
client/posts: remember offset when opening/closing bulk editor

### LINKED ISSUE #274: Bug: Clicking "Mass tag -> Start" or "Mass edit safety" resets current page number
Go to any page other than the first page -> start any of the mass edit actions -> page number resets to 1, so now you have to navigate back to where you wanted to mass edit.

Just adding it here for completeness. I'll probably fix this myself sometime, but I can't self assign.

neobooru:
Fixes #274

This does remove the `parameters.query === this._ctx.parameters.query` logic, but I couldn't figure out what it was meant to do? That said, if you want to keep it I could add it back.

### DIFF
--- a/client/js/views/posts_header_view.js
+++ b/client/js/views/posts_header_view.js
@@ -248,7 +248,7 @@ class PostsHeaderView extends events.EventTarget {
         let parameters = {query: this._queryInputNode.value};
         
         // convert falsy values to an empty string "" so that we can correctly compare with the current query
-        var prevQuery = this._ctx.parameters.query ? this._ctx.parameters.query : "";
+        const prevQuery = this._ctx.parameters.query ? this._ctx.parameters.query : "";
         parameters.offset = parameters.query === prevQuery ? this._ctx.parameters.offset : 0;
         if (this._bulkTagEditor && this._bulkTagEditor.opened) {
             parameters.tag = this._bulkTagEditor.value;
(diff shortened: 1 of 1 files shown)


rr-:
I think it could be about resetting the page number whenever you search for something new via the input field.

neobooru:
Yeah that seems to be the case. I'll debug some more and add it back.

client/js/views/posts_header_view.js
```diff
@@ -246,8 +246,10 @@ class PostsHeaderView extends events.EventTarget {
     _navigate() {
         this._autoCompleteControl.hide();
         let parameters = {query: this._queryInputNode.value};
-        parameters.offset = parameters.query === this._ctx.parameters.query ?
-            this._ctx.parameters.offset : 0;
+        
+        // convert falsy values to an empty string "" so that we can correctly compare with the current query
+        var prevQuery = this._ctx.parameters.query ? this._ctx.parameters.query : "";
```
rr-: `const`

neobooru:
Okay so the problem was that this._ctx.parameters.query would be `undefined` when there was no previous query. And `this._ctx.parameters.query === ""` returns `false`. This works around that by using the prevQuery variable which will be set to `""` when there are any falsy values (so null, undefined, "").

sgsunder:
Squashed and merged in eb49aea,  thanks!

PR #304, opened 2020-03-13
Migrate Elasticsearch functionality into Postgres

sgsunder:
Hey, @rr- , I would like you to review this before I merge it in.

This basically moves the reverse image search functionality that relied on Elasticsearch into Postgres, which was surprisingly easy. I test ran it on a copy of my personal instance (with about 4k images) and it seems to have similar performance to Elasticsearch as well, while being much more lightweight.

Only thing that feels a bit hacky to me is that I had to write raw SQL for the image search query. It seems like SQLAlchemy doesn't have a full implementation of Postgres' `unnest` function:
https://groups.google.com/forum/#!msg/sqlalchemy/thc9rla91gU/uFpXTAFKAwAJ

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -36,45 +36,31 @@ and Docker Compose (version 1.6.0 or greater) already installed.
 
 ### Running the Application
 
-1. Configurations for ElasticSearch:
+Download containers:
+```console
+user@host:szuru$ docker-compose pull
+```
 
-    You may need to raise the `vm.max_map_count`
-    parameter to at least `262144` in order for the
-    ElasticSearch container to function. Instructions
-    on how to do so are provided
-    [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-cli-run-prod-mode).
+For first run, it is recommended to start the database separately:
+```console
+user@host:szuru$ docker-compose up -d sql
+```
 
-2. Running the application
+To start all containers:
+```console
+user@host:szuru$ docker-compose up -d
+```
 
-    Download containers:
-    ```console
-    user@host:szuru$ docker-compose pull
-    ```
+To view/monitor the application logs:
+```console
+user@host:szuru$ docker-compose logs -f
+# (CTRL+C to exit)
+```
 
-    For first run, it is reccomended to start the databases seperately:
-    ```console
-    user@host:szuru$ docker-compose up -d elasticsearch
-    user@host:szuru$ docker-compose up -d sql
-    ```
-    Then wait approx. 2 minutes before starting all containers.
-    This gives time for the databases to initalize their storage
-    structure.
-
-    To start all containers:
-    ```console
-    user@host:szuru$ docker-compose up -d
-    ```
-
-    To view/monitor the application logs:
-    ```console
-    user@host:szuru$ docker-compose logs -f
-    # (CTRL+C to exit)
-    ```
-
-    To stop all containers:
-    ```console
-    user@host:szuru$ docker-compose down
-    ```
+To stop all containers:
+```console
+user@host:szuru$ docker-compose down
+```
 
 ### Additional Features
 
--- a/server/config.yaml.dist
+++ b/server/config.yaml.dist
@@ -144,9 +144,3 @@ privileges:
 ## usage: schema://user:password@host:port/database_name
 ## example: postgres://szuru:dog@localhost:5432/szuru_test
 #database:
-#elasticsearch: # used for reverse image search
-#    host: localhost
-#    port: 9200
-#    index: szurubooru
-#    user: szurubooru
-#    pass:
--- a/server/config.yaml.dist
+++ b/server/config.yaml.dist
@@ -21,10 +21,14 @@ thumbnails:
     post_width: 300
     post_height: 300
 
+# automatically convert animated GIF uploads to video formats
 convert:
    gif:
      to_webm: false
      to_mp4: false
+
+# allow posts to be uploaded even if some image processing errors occur
+allow_broken_uploads: false
 
 # used to send password reset e-mails
 smtp:
--- a/server/szurubooru/func/image_hash.py
+++ b/server/szurubooru/func/image_hash.py
@@ -2,8 +2,7 @@ import logging
 from io import BytesIO
 from datetime import datetime
 from typing import Any, Optional, Tuple, Set, List, Callable
-import elasticsearch
-import elasticsearch_dsl
+import math
 import numpy as np
 from PIL import Image
 from szurubooru import config, errors
@@ -24,30 +23,25 @@ N = 9
 P = None
 SAMPLE_WORDS = 16
 MAX_WORDS = 63
-ES_DOC_TYPE = 'image'
-ES_MAX_RESULTS = 100
+SIG_CHUNK_BITS = 32
+
+SIG_BASE = 2*N_LEVELS + 2
+SIG_CHUNK_WIDTH = int(SIG_CHUNK_BITS / math.log2(SIG_BASE))
+SIG_CHUNK_NUMS = 8*N*N / SIG_CHUNK_WIDTH
+assert 8*N*N % SIG_CHUNK_WIDTH == 0
 
 Window = Tuple[Tuple[float, float], Tuple[float, float]]
-NpMatrix = Any
-
-
-def get_session() -> elasticsearch.Elasticsearch:
-    extra_args = {}
-    if config.config['elasticsearch']['pass']:
-        extra_args['http_auth'] = (
-            config.config['elasticsearch']['user'],
-            config.config['elasticsearch']['pass'])
-        extra_args['scheme'] = 'https'
-        extra_args['port'] = 443
-    return elasticsearch.Elasticsearch([{
-        'host': config.config['elasticsearch']['host'],
-        'port': config.config['elasticsearch']['port'],
-    }], **extra_args)
+NpMatrix = np.ndarray
 
 
 def _preprocess_image(content: bytes) -> NpMatrix:
-    img = Image.open(BytesIO(content))
-    return np.asarray(img.convert('L'), dtype=np.uint8)
+    try:
+        img = Image.open(BytesIO(content))
+        return np.asarray(img.convert('L'), dtype=np.uint8)
+    except IOError:
+        raise errors.ProcessingError(
+            'Unable to generate a signature hash '
+            'for this image.')
 
 
 def _crop_image(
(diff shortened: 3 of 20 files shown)


doc/INSTALL.md
```diff
@@ -36,45 +36,31 @@ and Docker Compose (version 1.6.0 or greater) already installed.
...
-    To stop all containers:
-    ```console
-    user@host:szuru$ docker-compose down
-    ```
+Download containers:
+```console
+user@host:szuru$ docker-compose pull
+```
+
+For first run, it is reccomended to start the database seperately:
```
rr-: Typo: `reccommended` → `recommended`
Typo: `seperately` → `separately`

server/config.yaml.dist
```diff
@@ -21,11 +21,15 @@ thumbnails:
     post_width: 300
     post_height: 300
 
+# automatically convert animated GIF uploads to video formats
 convert:
    gif:
      to_webm: false
      to_mp4: false
 
+# allow posts to upload even if some image processing errors occur
```
rr-: Grammar: `to be uploaded`

server/szurubooru/tests/func/test_posts.py
```diff
@@ -964,3 +980,21 @@ def test_merge_posts_replaces_content(
     assert os.path.exists(source_path)
     assert os.path.exists(target_path1)
     assert not os.path.exists(target_path2)
+
+
+def test_search_by_image(
+        post_factory, config_injector, read_asset):
```
rr-: Can fit in a single line

server/szurubooru/tests/func/test_posts.py
```diff
@@ -964,3 +980,21 @@ def test_merge_posts_replaces_content(
...
+    config_injector({'allow_broken_uploads': False})
+    post = post_factory()
+    # db.session.flush()
```
rr-: No reason to commit in commented code

server/szurubooru/tests/func/test_posts.py
```diff
@@ -964,3 +980,21 @@ def test_merge_posts_replaces_content(
...
+    db.session.flush()
+
+    result1 = posts.search_by_image(read_asset('jpeg-similar.jpg'))
+    assert len(result1) == 1
+    result1_distance, result1_post = result1[0]
+    assert abs(result1_distance - 0.20599895341812172) < 1e-8
+    assert result1_post.post_id == post.post_id
+
+    result2 = posts.search_by_image(read_asset('png.png'))
+    assert not result2
```
rr-: Thanks for adding tests for this.

server/szurubooru/tests/func/test_posts.py
```diff
@@ -388,6 +387,7 @@ def test_update_post_content_to_existing_content(
             'post_height': 300,
         },
         'secret': 'test',
+        'allow_broken_uploads': False,
```
rr-: I'd separate this logic around preventing broken uploads into a separate pull request

server/szurubooru/func/posts.py
```diff
@@ -485,14 +478,61 @@ def test_sound(post: model.Post, content: bytes) -> None:
...
 
 
+def purge_post_signature(post: model.Post) -> None:
+    old_signature = (
+        db.session
+        .query(model.PostSignature)
+        .filter(model.PostSignature.post_id == post.post_id)
+        .one_or_none())
+    if old_signature:
+        db.session.delete(old_signature)
```
rr-: Maybe it could be shortened to `db.session.filter(…).delete()`?

server/szurubooru/tests/func/test_image_hash.py
```diff
@@ -1,37 +1,26 @@
...
-    })
+def test_signature_functions(read_asset, config_injector):
+    sig1 = image_hash.generate_signature(read_asset('jpeg.jpg'))
+    sig2 = image_hash.generate_signature(read_asset('jpeg-similar.jpg'))
 
-    if not image_hash.get_session().ping():
-        pytest.xfail(
-            'Unable to connect to ElasticSearch, '
-            'perhaps it is not available for this test?')
+    sig1_repaked = image_hash.unpack_signature(
```
rr-: `repacked`?

server/szurubooru/tests/func/test_image_hash.py
```diff
@@ -1,37 +1,26 @@
...
-    assert results_exact[0].distance == 0
-    assert results_similar[0].path == 'test'
-    assert results_similar[0].score == 17
-    assert abs(results_similar[0].distance - 0.20599895341812172) < 1e-8
+    words1 = image_hash.generate_words(sig1)
+    words2 = image_hash.generate_words(sig2)
+    words_match = 0
+    for word1, word2 in zip(words1, words2):
+        if word1 == word2:
+            words_match = words_match + 1
```
rr-: `words_match = sum(word1 == word2 for word1, word2 in zip(words1, words2))`

server/szurubooru/tests/func/test_image_hash.py
```diff
@@ -1,37 +1,26 @@
...
+    assert words_match == 17
```
rr-: I'd still test for distance being 0 if you try to compare a hash against itself

server/szurubooru/func/posts.py
```diff
@@ -768,38 +822,31 @@ def search_by_image_exact(image_content: bytes) -> Optional[model.Post]:
...
+    query_signature = image_hash.generate_signature(image_content)
+    query_words = image_hash.generate_words(query_signature)
+
+    dbquery = '''
+    SELECT s.post_id, s.signature, count(a.query) AS score
+    FROM post_signature AS s, unnest(s.words, :q) AS a(word, query)
+    WHERE a.word = a.query
+    GROUP BY s.post_id
+    ORDER BY score DESC LIMIT 100;
+    '''
```
rr-: I'm not familiar with `unnest`, could you elaborate a bit about it? Perhaps in a docstring :–] ?

server/szurubooru/func/image_hash.py
```diff
@@ -235,124 +228,21 @@ def _normalized_distance(
...
-@_safety_blanket(lambda: set())
-def get_all_paths() -> Set[str]:
-    search = (
-        elasticsearch_dsl.Search(
-            using=get_session(),
-            index=config.config['elasticsearch']['index'],
-            doc_type=ES_DOC_TYPE)
-        .source(['path']))
-    return set(h.path for h in search.scan())
+def pack_signature(signature: NpMatrix) -> bytes:
```
rr-: Could we have a docstring for this class?

server/szurubooru/func/image_hash.py
```diff
@@ -235,124 +228,21 @@ def _normalized_distance(
...
+    base = 2 * N_LEVELS + 1
+    coding_vector = np.flipud(SIG_BASE**np.arange(SIG_CHUNK_WIDTH))
+    return np.array([
+        np.dot(x, coding_vector) for x in
+        np.reshape(signature + N_LEVELS, (-1, SIG_CHUNK_WIDTH))
+    ]).astype(f'uint{SIG_CHUNK_BITS}').tobytes()
+
+
+def unpack_signature(packed: bytes) -> NpMatrix:
```
rr-: Could we have a docstring for this class?

server/szurubooru/func/posts.py
```diff
@@ -751,6 +803,8 @@ def transfer_flags(source_post_id: int, target_post_id: int) -> None:
     if replace_content:
         content = files.get(get_post_content_path(source_post))
         transfer_flags(source_post.post_id, target_post.post_id)
+        purge_post_signature(source_post)
+        purge_post_signature(target_post)
```
rr-: If it's about merging posts, maybe we could just reassign `.signature` objects rather than purging them?

server/szurubooru/func/posts.py
```diff
@@ -515,18 +555,30 @@ def update_post_content(post: model.Post, content: Optional[bytes]) -> None:
             and other_post.post_id != post.post_id:
         raise PostAlreadyUploadedError(other_post)
 
+    if update_signature:
+        purge_post_signature(post)
+        post.signature = generate_post_signature(post, content)
```
rr-: Maybe we could do it periodically in that background thread instead? Then again, it'd cause the db to stay inconsistent…

server/szurubooru/tests/func/test_posts.py
sgsunder: No problem!

server/szurubooru/func/posts.py
sgsunder: Seems to work.

server/szurubooru/func/posts.py
sgsunder: I tried various combinations of this and constantly run into issues with SQLAlchemy's 'ON DELETE CASCADE' implementation.

It sucks that we have to regenerate the signature, but I don't think users are merging often enough for this to be a major setback.

rr-: What if you flush inbetween?
Also that's fine but I see that you're only removing the signatures of both posts but not regenerating them, which means it'll stay null until the server gets restarted and repopulates missing hashes. Or am I missing something?

server/szurubooru/tests/func/test_posts.py
sgsunder: I think it's necessary for us to properly define the error handling characteristics before we merge in this PR. If really needed, we could break this off and merge it first. But if we are going to merge it anyway I would rather save on the effort. 🙃

server/szurubooru/func/posts.py
sgsunder: One nice aspect about the image uploads is that when I bulk upload variants of one image, the reverse search will immediately find a hit to the images just uploaded, allowing me to define a relation for them.

server/szurubooru/tests/func/test_posts.py
rr-: Yeah makes sense, I'd also like to avoid having to review it twice

server/szurubooru/func/posts.py
rr-: Yeah I'm totally fine with this approach, as long as it doesn't have unreasonably long run time.

server/szurubooru/func/posts.py
sgsunder: Flushing doesn't seem to help. I wouldn't be surprised to find out it's a bug in SQLAlchemy, cause signatures cascade delete on the `tests/model/test_post.py` unit test just fine.

The call to `update_post_content` just below will regenerate the hashes.

server/szurubooru/func/image_hash.py
sgsunder: Take a look, I think this should be sufficient?

server/szurubooru/func/posts.py
sgsunder: Wrote a small docstring for this, let me know if it makes sense. It's kinda tricky to verbalize without visual aids.

sgsunder:
I made a copy of my personal instance, which has a bit over 4k image posts. The performance I noticed was about the same. However, I have noticed that the main bottleneck in the reverse search was not the database query (to either Elasticsearch or Postgres), but the calculations done in NumPy.

I haven't done any sort of proper benchmarking, and I'm not sure how to approach doing that in the first place.

Something that I forgot to mention that will probably increase performance: Right now the SQL query retrieves the 100 best scoring posts for any post with a non-zero score. This basically mimics how the Elasticsearch query was set up.

I could add a clause such as `HAVING score > _num_` that would additionally filter off low scoring values, and put less load on the NumPy calculations. Only question is what should the minimum score be?

rr-:
Difficult to answer, let's keep it as is IMO. It's always possible to revise it later, for now let's merge getting rid of elasticsearch.

sgsunder:
I'm thinking of tagging the current commit as `2.3` before merging this, so that any users really attached to Elasticsearch for whatever reason can freeze their version. Thoughts?

rr-:
Sounds good

PR #307, opened 2020-03-15
server/func/tags: allow tags to have longer names

neobooru:
Because some series are called itai_no_wa_iya_nano_de_bougyoryoku_ni_kyokufuri_shitai_to_omoimasu.
128 seemed like the most logical number after 64. I guess 80 or something like that would suffice, but then we might have another PR in 4 months. Space-wise it doesn't actually matter.


sgsunder:
This looks fine to me.

@rr- was there a reason you only used 64 characters for tag names that I might be missing?

rr-:
No particular reason :) 64 seemed to me more than enough at the time, but I see how it can be a bit short, especially for anime-themed boards and the current anime naming schemes.

PR #310, opened 2020-04-02
Allow for fetching post content with youtube-dl

sgsunder:
- Allows for video-based posts to be created by using youtube-dl.  Now you can directly pass a URL to a site containing a video file (youtube, gfycat, etc) to generate a video post, instead of having to find a hotlink to the .mp4/.webm file (if one even exists).
- While convenient, there is potential for abuse of this feature, so a new permission "uploads:use_downloader" has been created, defaulting to power users and above.
- Default flag behavior has been changed globally (to partially support this). When using the web interface, the "loop" flag is set by default for all video posts. Users accessing the API directly can circumvent the defaults by ensuring that the `flag` parameter is set in the request.

### DIFF
--- a/server/szurubooru/func/net.py
+++ b/server/szurubooru/func/net.py
@@ -31,9 +31,12 @@ def download(url: str, use_video_downloader: bool = False) -> bytes:
 def _youtube_dl_wrapper(url: str) -> bytes:
     options = {
         'quiet': True,
+        'ignoreerrors': False,
         'format': 'webm/mp4',
         'logger': logger,
         'noplaylist': True,
+        'max_filesize': config.config['max_dl_filesize'],
+        'max_downloads': 1,
         'outtmpl': os.path.join(
             config.config['data_dir'],
             'temporary-uploads',
@@ -42,9 +45,19 @@ def _youtube_dl_wrapper(url: str) -> bytes:
     with YoutubeDL(options) as ydl:
         try:
             ydl_info = ydl.extract_info(url, download=True)
+            # need to confirm if download was skipped due to size
+            if ydl_info['filesize'] > config.config['max_dl_filesize']:
+                raise errors.DownloadTooLargeError(
+                    'Requested video too large (%d MB > %d MB)' % (
+                        ydl_info['filesize'] / 1.0e6,
+                        config.config['max_dl_filesize'] / 1.0e6))
             ydl_filename = ydl.prepare_filename(ydl_info)
         except YoutubeDLError as ex:
             raise errors.ThirdPartyError(
                 'Error downloading video %s (%s)' % (url, ex))
-    with open(ydl_filename, 'rb') as f:
-        return f.read()
+    try:
+        with open(ydl_filename, 'rb') as f:
+            return f.read()
+    except FileNotFoundError as ex:
+        raise errors.ThirdPartyError(
+            'Error downloading video %s' % (url))
--- a/server/szurubooru/func/net.py
+++ b/server/szurubooru/func/net.py
@@ -1,9 +1,17 @@
+import logging
 import urllib.request
-from szurubooru import config
-from szurubooru import errors
+import os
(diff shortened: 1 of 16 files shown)


sgsunder:
@rr- would you take a look at this before I merge it?

rr-:
I don't understand the rationale of removing the loop flag

server/szurubooru/func/net.py
```diff
@@ -1,16 +1,51 @@
...
     try:
         with urllib.request.urlopen(request) as handle:
-            return handle.read()
+            content = handle.read()
     except Exception as ex:
         raise errors.ProcessingError('Error downloading %s (%s)' % (url, ex))
+    if (use_video_downloader and
+            mime.get_mime_type(content) == 'application/octet-stream'):
+        return _youtube_dl_wrapper(url)
+    else:
```
rr-, line 28: Unnecessary `else`

server/szurubooru/func/net.py
```diff
@@ -1,16 +1,51 @@
...
+    }
+    with YoutubeDL(options) as ydl:
+        try:
+            ydl_info = ydl.extract_info(url, download=True)
+            ydl_filename = ydl.prepare_filename(ydl_info)
+        except YoutubeDLError as ex:
+            raise errors.ThirdPartyError(
+                'Error downloading video %s (%s)' % (url, ex))
+    with open(ydl_filename, 'rb') as f:
+        return f.read()
```
rr-, line 51: What about downloading 2 hour long documentaries and the max post file size error?
If we are going to support this, it'd also better not clog the server with downloading 4 gigabytes until it figures that it exceeded the limit of 10 mb :D

server/szurubooru/tests/func/test_net.py
```diff
@@ -47,3 +50,26 @@ def test_download(config_injector):
...
     actual_content = net.download(url)
     assert actual_content == expected_content
+
+
+def test_video_download(tmpdir, config_injector):
+    config_injector({
+        'user_agent': None,
+        'data_dir': str(tmpdir.mkdir('data'))
+    })
+    url = 'https://www.youtube.com/watch?v=C0DPdy98e4c'
```
rr-, line 60: I hoped for something fun :D

sgsunder:
If we want to keep the "loop" flag toggle available on client uploads, then we would need to add some additional client-side logic to determine when a link is a video type. I'm not really sure if that's feasible if the link is a youtube-dl link. (You would basically have to download it first before being able to tell if it's a video or image).

I think setting the loop flag default-on on video posts is reasonable, because most people are only going to put short GIF-like content for video posts. For edge cases, you can still edit the post manually after upload.

rr-:
>For edge cases, you can still edit the post manually after upload.

Oh, sorry, I thought you also removed it from the edit page for some reason. In that case it's all good.

I like the idea, but I'm concerned about how this plays with big uploads. I'm not convinced about power users only, cause a few 15 seconds long videos in 4k can easily waste a few hundred MB if not more, and that's a honest mistake, nothing malicious. This is something that can be iterated further, possibly after merging.

sgsunder:
Hmm, there's a `max_filesize` option for youtube-dl, I can quickly add that in before merging.

One thing to note is that I don't think the prior downloader had any limits, either, i.e. you could provide a hotlink to a large .mp4 file and it would download it too, so I might try and add a limit for the regular downloader as well.

rr-:
>One thing to note is that I don't think the prior downloader had any limits, either

Oh, gosh, you may be right. That's totally a security hole in my opinion :D

sgsunder:
Ok, I added a filesize check for at least the 'youtube-dl' downloads. I'm not sure how to approach implementing it in the regular downloader, so I'm leaving that as a todo for now.

rr-:
Sounds good, thanks. I mean, it was good to merge as-is and keep it as a todo too but it's nice that we already have it :)

PR #324, opened 2020-05-18
client/css: make add/remove button for mass tag larger

### LINKED ISSUE #322: Mass tag 'add' button is too hard to press, causing reduced tagging speed
When mass tagging would it be better if the 'add' button has a larger size, or at least the ability to open a post's details is disabled? I find that it's too hard to press since I have to 'aim' the cursor to avoid clicking on the image thumbnail by accident.

Somewhat related is I found that the mass tagging state is reset if you click on "Posts" in the top bar after viewing a post's details, but not when using the browser's back button.

sgsunder:
Fixes #322

### DIFF
--- a/client/css/post-list-view.styl
+++ b/client/css/post-list-view.styl
@@ -70,7 +70,7 @@
                         height: 1em
                         text-align: center
                         line-height: 1em
-                        font-size: 1.6em
+                        font-size: 2.2em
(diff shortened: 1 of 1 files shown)


sgsunder:
Merged in b037ce80c31962369c8e51060859ead6b48a1b92

PR #335, opened 2020-07-08
Add optional reCAPTCHA for registration

Kangaroux:
This added a recaptcha button to the registration page. The button only appears if recaptcha is enabled in the config and the user is not logged in.

I included a couple links in the config to the recaptcha docs. All that's needed to use it is to enable it in the config and provide the two tokens. By default the recaptcha is disabled.

```yaml
enable_recaptcha: yes
recaptcha_site_key: change-me
recaptcha_secret: change-me
```

## Captcha
image

## Captcha error message
image

### DIFF
--- a/client/html/index.htm
+++ b/client/html/index.htm
@@ -22,6 +22,7 @@
     <link rel='apple-touch-startup-image' href='img/apple-touch-startup-image-1668x2224.png' media='(min-device-width: 834px) and (max-device-width: 834px) and (-webkit-min-device-pixel-ratio: 2) and (orientation: portrait)'/>
     <link rel='apple-touch-startup-image' href='img/apple-touch-startup-image-2048x2732.png' media='(min-device-width: 1024px) and (max-device-width: 1024px) and (-webkit-min-device-pixel-ratio: 2) and (orientation: portrait)'/>
     <link rel='manifest' href='manifest.json'/>
+    <script src="https://www.google.com/recaptcha/api.js"></script>
 </head>
 <body>
     <div id='top-navigation-holder'></div>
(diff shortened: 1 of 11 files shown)


server/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN \
         alembic \
         "coloredlogs==5.0" \
         youtube-dl \
-    && apk --no-cache del py3-pip
+        requests
```
Kangaroux, line 27: Deleting pip caused the requests library to not be importable.

client/html/index.htm
```diff
@@ -22,6 +22,7 @@
     <link rel='apple-touch-startup-image' href='img/apple-touch-startup-image-1668x2224.png' media='(min-device-width: 834px) and (max-device-width: 834px) and (-webkit-min-device-pixel-ratio: 2) and (orientation: portrait)'/>
     <link rel='apple-touch-startup-image' href='img/apple-touch-startup-image-2048x2732.png' media='(min-device-width: 1024px) and (max-device-width: 1024px) and (-webkit-min-device-pixel-ratio: 2) and (orientation: portrait)'/>
     <link rel='manifest' href='manifest.json'/>
+    <script src="https://www.google.com/recaptcha/api.js"></script>
```
ReAnzu, line 25: I would prefer that we don't load the reCAPTCHA js library if the feature is disabled.

sgsunder:
Please address my review comments

Kangaroux:
@sgsunder Check that you submitted the review, I don't see anything

client/html/index.htm
Kangaroux: Not entirely sure how to do that on the base template to be honest.

sgsunder:
Regarding the Docker Cloud CI failure: Make sure that any changes you make to `server/Dockerfile` is mirrored in `server/Dockerfile.test`

PR #349, opened 2020-09-19
server/pools: serialize pools as micro resource within post resources

### LINKED ISSUE #348: Feature request: change the way "pools" work inside a post resource
Today I started using pools and I created a pool with ~200 posts in it.
When I request a `/api/posts` from the server I get roughly the following:
```json5
{
    "query": "",
    "offset": 0,
    "limit": 100,
    "total": 1000,
    "results": [
        {
            "id": 1,
            // Removed uninteresting fields
            "pools": [
                {
                    "id": 1,
                    "names": [
                        "test_pool_1"
                    ],
                    // Removed uninteresting fields
                    "posts": [
                        {
                            "id": 1,
                            "thumbnailUrl": "https://..."
                        },
                        {
                            "id": 2,
                            "thumbnailUrl": "https://..."
                        },
                        // This is repeated for ~200 times
                    ]
                }
            ],
        },
        {
            // Another post resources, with inside it another pools list with ~200 items
        },
        {
            // Another post resources, with inside it another pools list with ~200 items
        },
        {
            // Another post resources, with inside it another pools list with ~200 items
        }
    ]
}
```

Because of this I end up with a 4MB `/api/posts` response (usually only 300KB), which will only grow over time.

A great solution would be to add a micro pool resource, like we have for tags and posts.

This micro pool could look like this:
```json5
{
    "id":           <id>,
    "names":        <names>,
    "category":     <category>,
    "postCount":    <post-count>,
    "description":  <description>
}
```

I'd love to hear any suggestions or alternative ideas.

sgsunder:
Fixes #348

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -99,6 +99,7 @@
    - [Micro post](#micro-post)
    - [Pool category](#pool-category)
    - [Pool](#pool)
+   - [Micro pool](#micro-pool)
    - [Note](#note)
    - [Comment](#comment)
    - [Snapshot](#snapshot)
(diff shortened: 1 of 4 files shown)


sgsunder:
@neobooru I don't have a large pools dataset to test this on. Can you check to see if this works for you?

neobooru:
The response body which was 4MB is 170KB when using the new micro-pool resource. I'd say that it works like a charm.

PR #350, opened 2020-09-19
Add tag category ordering feature

neobooru:
How does this look?

This sorts by `order -> isDefault -> categoryName -> tagName`
This doesn't change the default behavior (sorted a-z and default tags last) because order = 1 by default.

Didn't write any tests for this yet, but seems to work fine in my test setup.

image


server/szurubooru/api/tag_category_api.py
```diff
@@ -73,6 +73,11 @@ def update_tag_category(
         tag_categories.update_category_color(
             category, ctx.get_param_as_string("color")
         )
+    if ctx.has_param("order"):
+        auth.verify_privilege(ctx.user, "tag_categories:edit:order")
+        tag_categories.update_category_order(
+            category, ctx.get_param_as_int("order")
+        )
```
sgsunder, line 80: Copy this code diff into the `create_tag_category` method above this one. Current code is unable to set the order upon category creation, only when editing existing ones.

server/szurubooru/func/tag_categories.py
```diff
@@ -117,6 +121,11 @@ def update_category_color(category: model.TagCategory, color: str) -> None:
     category.color = color
 
 
+def update_category_order(category: model.TagCategory, order: int) -> None:
+    assert category
+    category.order = order
```
sgsunder, line 126: I think we should verify that the given `order` is a positive integer here, and maybe add an upper bound, too. (Unless we really want the full range of ints to be available?)

Would recommend creating an `InvalidTagCategoryOrderError` exception to go with this as well.

client/js/models/tag_category.js
```diff
@@ -58,16 +68,19 @@ class TagCategory extends events.EventTarget {
...
         }
 
         let promise = this._origName
             ? api.put(
-                  uri.formatApiLink("tag-category", this._origName),
-                  detail
-              )
+                uri.formatApiLink("tag-category", this._origName),
+                detail
+            )
```
sgsunder, line 83: Should install and enable [pre-commit](https://pre-commit.com/) to ensure everything is of a consistent style

sgsunder (changes requested):
Please see the review comments.

Additionally, we should document the tag category API changes in `API.md`

server/szurubooru/func/tag_categories.py
neobooru: I see no reason to restrict it to only positive integers, I could definitely see someone adding a negative ordered category instead of updating the order of all other categories. I'll check what happens when you go over the INT_MAX (not sure if that is a problem for the database), but I generally don't think we should restrict things without a good/technological reason.

client/js/models/tag_category.js
neobooru: I believe this is the correct formatting, and that it was wrong before. pre-commit also formats it like this.

server/szurubooru/func/tag_categories.py
sgsunder: Ok, sounds reasonable

sgsunder (approved):

sgsunder:
Made a PR on your branch to fix the failing tests: https://github.com/neobooru/szurubooru/pull/10

Issue #216, opened 2018-12-27
Unable to upload: PermissionError: [Errno 13] Permission denied: '/data/temporary-uploads'


WychWitch:
Hello! I'm sorry for always posting on here but this time I'm *extremely* close to getting this working.

I recently gave up trying to run this on my raspberry pi 1 and decided to test it on my pi 3, and after configuring everything for armv7 it finally worked! However, I'm running into a strange permission error.

<details>
<summary>
Here's the full stack trace of an error, and this pops up every time I try to upload a picture.
</summary>

```
client_1         | 2018/12/27 15:19:56 [warn] 18#18: *1 a client request body is buffered to a temporary file /var/tmp/nginx/client_body/0000000001, client: 192.168.1.9, server: , request: "POST /api/uploads HTTP/1.1", host: "192.168.1.28:8080", referrer: "http://192.168.1.28:8080/upload"
api_1            | [2018-12-27 15:19:57] szurubooru.middleware.request_logger POST /uploads (user=Testy-chan, queries=1)
api_1            | [2018-12-27 15:19:57] waitress Exception when serving /uploads
api_1            | Traceback (most recent call last):
api_1            |   File "/usr/local/lib/python3.6/site-packages/waitress/channel.py", line 338, in service
api_1            |     task.service()
api_1            |   File "/usr/local/lib/python3.6/site-packages/waitress/task.py", line 169, in service
api_1            |     self.execute()
api_1            |   File "/usr/local/lib/python3.6/site-packages/waitress/task.py", line 399, in execute
api_1            |     app_iter = self.channel.server.application(env, start_response)
api_1            |   File "/opt/app/szurubooru/rest/app.py", line 99, in application
api_1            |     response = handler(ctx, match.groupdict())
api_1            |   File "/opt/app/szurubooru/api/upload_api.py", line 11, in create_temporary_file
api_1            |     token = file_uploads.save(content)
api_1            |   File "/opt/app/szurubooru/func/file_uploads.py", line 29, in save
api_1            |     files.save(path, content)
api_1            |   File "/opt/app/szurubooru/func/files.py", line 40, in save
api_1            |     os.makedirs(os.path.dirname(full_path), exist_ok=True)
api_1            |   File "/usr/local/lib/python3.6/os.py", line 220, in makedirs
api_1            |     mkdir(name, mode)
api_1            | PermissionError: [Errno 13] Permission denied: '/data/temporary-uploads'
```

</details>

<details>
<summary>
[...]

PR #366, opened 2020-12-02
Docker: add installer.sh

ShockDW:
This pull request is meant to provide a user friendly install.sh for deploying Szurubooru.

Features:
- Prompts user for config values, then deploys Szurubooru automatically
- Performs the necessary ownership change of the local /data/ mount point to avoid problems uploading images #216 
- Checks host architecture and automates the changes outlined in readme.md and kaijuagenda's rpi3 fork for ARM hosts

I'm a big fan of documentation, so I tried my best to detail the purpose of the various operations and functions via comments. Naturally, these can be changed or removed if found to be too verbose or unclear.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -15,13 +15,8 @@
 # Check for root to ensure the accessibility of the Docker daemon, and to 
 # allow mount point permissions to be set after building.
 
-if [[ $EUID -ne 0 ]]; then
-    echo "Sorry, but this script needs root to properly function.  Please try again using sudo ./install.sh" 1>&2
-    exit 1
-fi
-
-echo "Welcome to Szurubooru.  This install script will ask you to set a few 
-echo "config values, then get szurubooru up and running."
+echo "Welcome to Szurubooru.  This install script will ask you to"
+echo "set a few config values, then get szurubooru up and running."
 
 # Check to ensure Docker and Docker-Compose are both installed before proceeding
 if ! command -v docker &> /dev/null || ! command -v docker-compose &> /dev/null; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,157 @@
...
+#
+# * Modifies docker-compose.yml rather than relying on a static alternative file to hopefully
+#   avoid the need for parallel maintenance of two files.
+#############################################################################################
+
+# Check for root to ensure the accessibility of the Docker daemon, and to 
+# allow mount point permissions to be set after building.
+
+if [[ $EUID -ne 0 ]]; then
+    echo "Sorry, but this script needs root to properly function.  Please try again using sudo ./install.sh" 1>&2
```
sgsunder, line 19: You don't always need root to run docker, see:

https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

install.sh
```diff
@@ -0,0 +1,157 @@
...
+########################################################################################
+
+if [[ "$(uname -m)" == 'a'* ]]; then
+    echo "ARM architecture detected.  Modfying docker-compose.yml accordingly."
+    sed -zi "s|image: szurubooru/server:latest|build:\n      context: ./server|; \     # Modify services/server to build locally instead of pulling from DockerHub
+             s|image: szurubooru/client:latest|build:\n      context: ./client|; \     # Modify services/client to build locally instead of pulling from DockerHub
+             s|depends_on:\n      - sql|depends_on:\n      - sql\n      - elasticsearch|; \ # Add dependency for alternate arm compatible elasticsearch defined below.
+             s|POSTGRES_HOST: sql|POSTGRES_HOST: sql\n      ESEARCH_HOST: elasticsearch|" \ # Point Postgres to the alternate elasticsearch container
+            docker-compose.yml
+    # "Add a new section to define the alternate elasticsearch service
```
sgsunder, line 47: You can drop this part, we haven't used elasticsearch in quite a while

sgsunder (changes requested):

sgsunder:
Thank you! This will probably help new users a lot

install.sh
ShockDW: Good point!  Removed Root check.

install.sh
ShockDW: Removed. :)

ShockDW:
My pleasure!  I'm fairly new to Python, so I'm just glad to have found a way to contribute.  Hopefully I can pitch in some more in the future.  

All suggested changes have been made, along with a few updates to output as well as a few fixes for flaws I noticed while testing said changes.  Please let me know if you notice anything else out of place, and thanks for the review!

install.sh
```diff
@@ -0,0 +1,140 @@
...
+#
+# * Modifies docker-compose.yml rather than relying on a static alternative file to hopefully
+#   avoid the need for parallel maintenance of two files.
+#############################################################################################
+
+echo "Welcome to Szurubooru.  This install script will ask you to"
+echo "set a few config values, then get szurubooru up and running."
+
+# Check to ensure Docker and Docker-Compose are both installed before proceeding
+if ! command -v docker &> /dev/null || ! command -v docker-compose &> /dev/null; then
```
Jazzepi, line 19: id -Gn | grep -i docker
^ will return 0 when the currently executing user in the docker group. You can unfortunately successfully run `docker` even if you don't have access to the socket it's talking to.

Additionally you could run `docker ps` to see if they can actually talk to their docker socket. It'll return 1 if they can't.

ShockDW: Excellent information!

This section was only really meant to serve as an install check for Docker and Docker-Compose were installed, but I've added an additional check on `docker ps`'s  return status to ensure the current user can access the daemon/socket.  Thanks Jazzepi!

Jazzepi: No problem :) I love what you did with the script, so kudos on that.

sgsunder (approved):

sgsunder:
I'll merge this in, but I think I will make further refinements later on to streamline this

PR #374, opened 2021-01-19
client/posts: provide link for E621 image search

SleepingCrows:
Fork of danbooru, same URL structure. Covers content not found on IQDB or Danbooru due to subject matter.

May be prudent in the future to make this functionality editable via the database, instead of being determined from within the TPL.


sgsunder:
I'm hesitant to add niche/NSFW sites because I want the base install to be usable by everyone. I'd recommend customizing the codebase yourself for people who want that level of customization.

> May be prudent in the future to make this functionality editable via the database

This might be an interesting idea to implement, though.

SleepingCrows:
Both i get, though if NSFW is the major issue, e926 doesn't return that content in search.

PR #375, opened 2021-02-03
client/posts: add post description

neobooru:
I've had this sitting in some far away directory for some time now, because I didn't like the design and the code.
I guess the best thing to do now is just ask some people for their opinion and/or improvements.

A few things I don't like/am not sure how to feel about:
- I am not sure if I like the padding the H1, H2, H3 tags add to the post description. It leaves an awful lot of empty space around the H1 tags (see first photo).
- I don't know whether I should show the `<h2>Description</h2>` on the edit page. (see 3rd and 4th photo).
- The code is inlined in `client/html/post_main.tpl`. Not sure if this is good or bad. (Reason for this is the extra div/h2 tags etc, which are just a pain to create in pure JS).

If you have other thoughts and/or tips please do share those too.

## Todo
- [ ] Post description edit permissions
- [ ] Idk?

## Images
[1. Post description](https://user-images.githubusercontent.com/50623835/106788262-04921a00-6651-11eb-9fcc-a677a5affc8d.png)
[2. Edit post description](https://user-images.githubusercontent.com/50623835/106788261-03f98380-6651-11eb-827f-29f61f5f225c.png)
[3. Edit post description resized](https://user-images.githubusercontent.com/50623835/106788259-0360ed00-6651-11eb-9634-edffd6aa086d.png)
[4. Edit post description (without H2)](https://user-images.githubusercontent.com/50623835/106789195-3b1c6480-6652-11eb-8d8b-a5c66be2327a.png)

### DIFF
--- a/client/css/post-main-view.styl
+++ b/client/css/post-main-view.styl
@@ -40,10 +40,18 @@
         width: 100%
 
     .post-container
-        margin-bottom: 2em
+        margin-bottom: 0.6em
 
         .post-content
             margin: 0
+
+    .description-container
+        background: $top-navigation-color
+        padding: 0.6em
+
+    #post-description
+        resize: vertical
+        min-height: 200px
 
 @media (max-width: 800px)
     .post-view
--- a/client/html/post_main.tpl
+++ b/client/html/post_main.tpl
@@ -54,6 +54,18 @@
     <div class='content'>
         <div class='post-container'></div>
 
+        <% if (ctx.editMode) { %>
+            <h2>Description</h2>
+            <%= ctx.makeTextarea({
+                id: 'post-description',
+                value: ctx.post.description,
+            }) %>
+        <% } else if (ctx.post.description != undefined) { %>
+            <div class='description-container'>
+                <%= ctx.makeMarkdown(ctx.post.description) %>
+            </div>
+        <% } %>
+
         <% if (ctx.canListComments) { %>
             <div class='comments-container'></div>
         <% } %>
--- a/client/js/controllers/post_main_controller.js
+++ b/client/js/controllers/post_main_controller.js
@@ -187,6 +187,9 @@ class PostMainController extends BasePostController {
         if (e.detail.source !== undefined) {
             post.source = e.detail.source;
         }
+        if (e.detail.description !== undefined) {
+            post.description = e.detail.description;
+        }
         post.save().then(
             () => {
                 this._view.sidebarControl.showSuccess("Post saved.");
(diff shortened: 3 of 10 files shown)


dankles:
> I am not sure if I like the padding the H1, H2, H3 tags add to the post description. It leaves an awful lot of empty space around the H1 tags (see first photo).

I agree that the margin on some of the header tags is very heavy. I've noticed this is also the same in tag descriptions, comments and other fields that support markdown. They all appear to inherit the same amount of margin. So if you left it as-is, it seems to be in line with the styling of other markdown supported text areas. 

> I don't know whether I should show the `<h2>Description</h2>` on the edit page. (see 3rd and 4th photo).

The `<h2>Description</h2>` on the edit page may not be needed if there is a `placeholder` attribute, describing what the textarea is for?

neobooru:
> The `<h2>Description</h2>` on the edit page may not be needed if there is a `placeholder` attribute, describing what the textarea is for?

That is a good idea, that would look something like [this](https://user-images.githubusercontent.com/50623835/107288445-c12a1800-6a63-11eb-946a-0cdf40e003d4.png). I could also change the text to something like "post description" if you want. I started the sentence with a lowercase because szurubooru does that (nearly) everywhere.

Do we also want the textarea to auto-resize to fit it's content? E.g. be small when only the placeholder is there, and larger when there's some content. Possibly with a max-height, or none at all?

dankles:
> That is a good idea, that would look something like this. I could also change the text to something like "post description" if you want. I started the sentence with a lowercase because szurubooru does that (nearly) everywhere.

The longer text is probably better, given it's a placeholder and not actually taking up any screen space other than the textarea.

> Do we also want the textarea to auto-resize to fit it's content? E.g. be small when only the placeholder is there, and larger when there's some content. Possibly with a max-height, or none at all?

I think you're talking about the edit page? On the edit page, I think resizing is ideal, given it's only comments that might get pushed down to fit it.

po5:
Dark theme styling in 177872fc13e1a5d834162a39b74b0014133d7f55

> Todo: Post description edit permissions

Done in 62469c09510d32fb23174f0c236e0a6fbfc91eef + 7d7e054b3f0cb1c30c2c2b2d205840ed113eaa82

Some oversights fixed in 106d1bfceb319de5be032c0d9ec0d442f1805194 + 72476c24eb289b8428a2f669c2ea6ea7b8bddf68 + 6eef29fc226784e98b2ccf72b4ced2544955db39

milezzz:
Hey this code works great, I see no errors. Really should be default behaviour and the SEO benefits will be huge. On that note, would be absolutely amazing if we could get this description into the meta of the /post pages now.

PR #387, opened 2021-04-09
Nginx added variable, to easier scale szuru.

calvinthefreak:
I made this changes with 4 Containers on my own instance and got up to 5K API requests/Sec on an Intel Celeron J4105.

Maybe you can change the stuff, I made.

### DIFF
--- a/client/nginx.conf.docker
+++ b/client/nginx.conf.docker
@@ -20,6 +20,7 @@ http {
     keepalive_timeout 65;
 
     upstream backend {
+        #Add more Container IPs/Aliases here.
         server __BACKEND__:6666;
     }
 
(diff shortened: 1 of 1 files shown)


sgsunder:
IMO, it would be better for users to use a proper load balancer like HAProxy or traefik (which integrates with Docker scaling) to do this, instead of using the Nginx container

calvinthefreak:
Hmm... I think Nginx would be a good option, because you can easily integrate some Firewall and Ratelimiting at the Edge.
That maybe is a good Idea IMO.

But I noticed, that LB over multiple Server Dockers is more performant. Especially if you Deploy the Setup over multiple Servers with Shared Storage.

sgsunder:
Yeah, but the way you have it right now stops you from doing it programatically, you have to manually edit the nginx configuration to add every single container. 

I'm not familiar with it, but Docker has the means to scale to running multiple containers for one service, and some load balancers can hook into the Docker API and automatically route HTTP calls across them.

> Firewall and Ratelimiting at the Edge

Both haproxy and traefik can do this.

calvinthefreak:
> Both haproxy and traefik can do this.
I will try out traefik and test it soon.

PR #388, opened 2021-04-09
Client CSS: Dark mode contrast fixes

robobenklein:
Before:

image

After:

image

Changes:
 - Above fix for page headers in dark mode
 - Changes the tag input bg highlighting to a CSS animation in order to better maintain readability of the tag name
 - In dark mode uses an outline box-shadow instead of bg highlight for post nav buttons
 - Darkens darktheme.message.success bg for readability contrast

### DIFF
--- a/client/css/core-general.styl
+++ b/client/css/core-general.styl
@@ -240,6 +240,9 @@ nav
                 background: $focused-tab-background-color-darktheme
     &#top-navigation
         background: $top-navigation-color-darktheme
+        ul
+            #mobile-navigation-toggle
+                color: $text-color-darktheme
 
 a .access-key
     text-decoration: underline
--- a/client/css/core-general.styl
+++ b/client/css/core-general.styl
@@ -275,7 +275,7 @@ a .access-key
             background: darken($message-error-background-color, 60%)
         &.success
             border: 1px solid darken($message-success-border-color, 30%)
-            background: darken($message-success-background-color, 60%)
+            background: darken($message-success-background-color, 80%)
 
 .thumbnail
     /*background-image: attr(data-src url)*/ /* not available yet */
--- a/client/css/pager.styl
+++ b/client/css/pager.styl
@@ -22,7 +22,7 @@
                 z-index: 1
             span
                 position: relative
-                background: white
+                background: $window-color
                 padding: 0 1em
                 z-index: 2
 
@@ -31,3 +31,5 @@
         .page-header
             &:before
                 background: $top-navigation-color-darktheme
+            span
+                background: $window-color-darktheme
--- a/client/css/post-main-view.styl
+++ b/client/css/post-main-view.styl
@@ -27,7 +27,7 @@
                     padding: 0.3em 0
                     text-align: center
                     vertical-align: middle
-                    transition: background 0.2s linear
+                    transition: background 0.2s linear, box-shadow 0.2s linear
                     &:not(.inactive):hover
                         background: lighten($main-color, 90%)
                 i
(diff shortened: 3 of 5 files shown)


neobooru:
For what it's worth, I personally feel like using the page bg color looks better. It also looks like that was the behavior of the light theme.

I also noticed some issues in the dark theme. I'll add them below, maybe you could fix those too? If not then I'll move them to their own issue.

On the post view page (when hovering the arrows)
image

When adding a tag. This isn't exactly an issue but imo it's just a tad too bright
image

When saving a post. The contrast is a bit too low, probably because the green is too bright.
image

Actually, only the first one is an actual issue. The rest are just me nitpicking.

robobenklein:
I agree with all of the issues you presented, I'll try to add all the suggestions in to this PR.

robobenklein:
I could not find a color or brightness that looked good against both the link color and the background, so I'll propose this instead:

GIFrecord_2021-04-12_185158

robobenklein:
While I'm changing it, might as well make it a little more interesting. (I was thinking about the case where a tag's color might be similar to one of the tag modification colors, so this should help do the job of attention-grabbing and also does not obscure a similar-colored tag for the whole transition.)

https://user-images.githubusercontent.com/5423266/114479227-16061e00-9bce-11eb-8529-e3190f542e22.mp4

https://user-images.githubusercontent.com/5423266/114479327-549bd880-9bce-11eb-9e41-76152f8441b3.mp4

This also revealed a bug where "new" tags that have implications did not have their "new" class assigned when added to compact-tags.

https://user-images.githubusercontent.com/5423266/114479884-62059280-9bcf-11eb-9348-986048c5cb7c.mp4

If the animation is too fast and we want to slow it down we'll have to ensure the JS keeps the class assigned for at least as long as the transition plays. (Looks like 3 seconds right now to me)

robobenklein:
@neobooru does that help with the issue of it being too bright as well? If not it's not hard to replicate the same backgrounds under `.darktheme` with `darken(c, 50%)`

neobooru:
> 
> 
> @neobooru does that help with the issue of it being too bright as well? If not it's not hard to replicate the same backgrounds under `.darktheme` with `darken(c, 50%)`

It looks good however you did it in your branch, imo.

The other changes are fine too I guess, but in the end that's up to the owners of this repo to decide. 
If you want to keep the old hover behavior you could also try `$active-tab-background-color-darktheme` as :hover bg.
(Though if we end up using that it'd probably to be cleaner to add a new variable)

The "new" and "duplicate" classname fadeout is handeled inside the `_fadeOutListItemNodeStatus` function. This function exists in both `pool_input_control.js` and `tag_input_control.js`.

robobenklein:
If the animations are accepted I can work on adding those to the other list input controls as well.

sgsunder:
I'm not a big fan of the sweep animation for new tags. I find the existing animation to be eye-catching enough without being distracting like the proposed one. Not really worried about the possibility of bad contrast with tag text because it's a transient animation regardless, so it wouldn't obstruct text for any meaningful length of time.

Then again, I'm not a UX designer, so if it makes a genuine improvement in accessibility or most people want it, I'm okay with it.

I like all the other changes made here though.

robobenklein:
image

`implication` still looks kinda ugly, taking suggestions

Might try something combining saturation and lightness stylus functions

robobenklein:
Anything else left to do on this PR for merge?

@sgsunder I took the sweep animation out of this PR, currently:
GIFrecord_2021-07-04_100123

Haven't noticed other visual problems to fix after I've had these changes live on multiple sites since, though if anyone else sees something else to change let me know.

neobooru:
I am still not a huge fan of having different animations for the dark theme (the nav buttons), but it is certainly an improvement over the previous look. All the other stuff looks good to, so I'll squash+merge this.

robobenklein:
@neobooru should I spend more time trying to find a color change only solution to the nav button contrast instead?

neobooru:
> 
> 
> @neobooru should I spend more time trying to find a color change only solution to the nav button contrast instead?

I wouldn't spend more time on this, it's good enough. Though if you ever find a better solution I'll happily merge it.

PR #390, opened 2021-04-10
Fix #385

### LINKED ISSUE #385: Problems with escaping periods in tags
Related to #97
As far as I can tell, tags with periods in them aren't escaped properly.
A tag that is just `..` ends up breaking completely, and will link to the base domain.
Video:

https://user-images.githubusercontent.com/22437119/112930722-ac671980-90cf-11eb-9adb-a87f7c97407d.mp4

neobooru:
Currently this does not fix the issue which arises when you go to view/edit a tag that is just dots, e.g `.` and `..`

This is probably because `.` means 'this directory' and `..` means 'parent directory'. I don't think having a tag like `..` is realistic, but feel free to correct me. Maybe escaping the dot as `%2E` would work, but [apparently this behavior differs between browsers.](https://stackoverflow.com/a/3857067)

This fails:
```sh
# View/edit tag with just dots
http://localhost/tag/.
http://localhost/tag/..

# This does work on my prod server, more on that below
http://localhost/tag/...
# etc
```

This works:
```sh
# Search for tag '...' escaped as '\.\.\.'
http://localhost/posts/query=%5C.%5C.%5C.
# Search for tag 'test...' escaped as 'test\.\.\.'
http://localhost/posts/query=test%5C.%5C.%5C.
```

However, some things work fine on my prod server, but not on my test setup.
```sh
# Works on prod, not on test. But I suspect that my test server is not configured correctly.
http://localhost/tag/test..
http://localhost/tag/test...
http://localhost/tag/...
```

Prod: linux with nginx (not docker)
Test: windows with nginx (not docker)

The difference is that `query=xxx` uses an escaped input, whereas `/tag/xxx` does NOT use an escaped input. 
On my production server `/tag/.../` works, but doesn't work with `.` and `..`, probably because `.` and `..` have a special meaning.

TL;DR:
- The `escapeTagName` function now also escapes dots, which fixes the issue illustrated in #385.
- Supporting `.` and `..` as a tag names seems unnecessarily complex, because of the special meaning of these dots.
- `/tag/***` does not use an escaped input. On my prod server this works fine, on my test server it does not. On my test server it seems to ignore the trailing dots, e.g. `test..test..` is recognized as `test..test`. In Fernsicles' video it seems to behave the same as on my prod server, aka the expected behavior. So maybe my test setup is just wrongly configured?

### DIFF
--- a/client/html/post_readonly_sidebar.tpl
+++ b/client/html/post_readonly_sidebar.tpl
@@ -92,7 +92,7 @@
                             --></a><!--
                         --><% } %><!--
                         --><% if (ctx.canListPosts) { %><!--
-                            --><a href='<%- ctx.formatClientLink('posts', {query: ctx.escapeColons(tag.names[0])}) %>' class='<%= ctx.makeCssName(tag.category, 'tag') %>'><!--
+                            --><a href='<%- ctx.formatClientLink('posts', {query: ctx.escapeTagName(tag.names[0])}) %>' class='<%= ctx.makeCssName(tag.category, 'tag') %>'><!--
                         --><% } %><!--
                             --><%- ctx.getPrettyName(tag.names[0]) %>&#32;<!--
                         --><% if (ctx.canListPosts) { %><!--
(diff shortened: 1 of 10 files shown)


neobooru:
It looks like `escapeTagName` and `escapeSearchTerm` have almost the same functionality. 
The only difference is that, given the input string `sort:random` they produce the following output:
escapeTagName = `sort\:random` 
escapeSearchTerm = `sort:random` 

However this functionality is currently not used. This would be used when `sort:ran` would be auto-completed to `sort:random`, but at this moment this doesn't happen.

sgsunder:
Merged, thanks!

PR #394, opened 2021-04-20
Suppressed the use of SQLAlchemy 1.4

nothink:
If the current requirements.txt is used and deployed in the on-premises environment, SQLAlchemy 1.4 series will be introduced and the test will not pass due to compatibility issues.

Therefore, we need to suppress the use of SQLAlchemy 1.4.


Dante4:
I guess it's stuck on sqlalchemy 1.3?

PR #400, opened 2021-05-08
Improve uploading process

Ruin0x11:
I wanted to fix two minor things about the upload page:

1. If you have a large number of files to upload but one of the files early on has an upload error, it will prevent all the files below it from being uploaded. For upload jobs with many files, it's really annoying to have to manually intervene after each error. I changed the way errors are handled so that if one post has an error the rest will continue uploading, and all the posts with errors will remain afterwards. (This could be made configurable if the behavior is actually undesirable.)
2. Some kinds of images like screenshots can naturally be similar to one another, but there is no way to tell the uploader that you don't care about similarity when uploading. I added a new option, "Always upload similar", which will skip the confirmation when a similar image is detected.

With these two enhancements it should be possible to send a large upload job and guarantee that all the images that are both of a supported format and not exact duplicates will be uploaded.

### DIFF
--- a/client/css/post-upload.styl
+++ b/client/css/post-upload.styl
@@ -13,8 +13,10 @@ $cancel-button-color = tomato
 
         &.inactive input[type=submit],
         &.inactive .skip-duplicates
+        &.inactive .always-upload-similar
         &.uploading input[type=submit],
         &.uploading .skip-duplicates,
+        &.uploading .always-upload-similar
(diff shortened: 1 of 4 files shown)


sgsunder:
Merged, thanks!

If you make any other PRs in the future, please make sure to enable and run pre-commit to ensure your code meets the project style

PR #402, opened 2021-05-09
Improve compilation speed for development builds

Ruin0x11:
1. Use watchify for incremental builds instead of naive polling (`npm run watch`). Speeds up the time between changes by ~17x.
2. When running `npm run watch`, set the environment as `development`, and open a websocket for live reloading. The event for live reloading is sent when any template/CSS/JS is changed. Can be disabled by passing `--no-live-reload` to `build.js`.
3. Better formatting of errors during incremental compilation.


neobooru:
Looks good.
I do have a small thing to nitpick about though. When you build the site for production (aka simple `npm run build`) the live-reloading code is still embedded into the app. I know something like webpack could remove this, but we aren't using that at the moment.
I assume it would be rather hard to remove those lines from the production build using our current build system, but I'm just asking to make sure.

Ruin0x11:
Maybe we should create a template that holds `main.js` and autogenerate it like with the HTML templates? It might make the code for that less clean though.

neobooru:
Thanks for your work 👍 
I squashed it to avoid adding the `watch.sh` to our history.

Issue #380, opened 2021-03-13
Feature request: Pool order and pool browsing implementation


404Fox:
This is a request for the rest of pools implementation. It doesn't seem there are any forks working on this right now.

See #321 for the initial PR adding pools. Quotes come from that post.

### Pool order is not searchable/enforced
>The order the posts are added in is saved to the database but it isn't used yet. The pool:id named token makes no assumptions about sorting, and it seems like the sort token feature only works on columns, so it seems difficult to make a `sort:pool:id` token since a single post belonging to many pools can be ordered differently depending on the pool searched for.

I don't know about the relationship of post-pool tables, but I'd assume the post doesn't need to be aware of its position in a pool? The order of the posts should be derived from the pool.

### Pool order cannot be modified in GUI
Draggable thumbnail GUI on the pool page. Or instead of dragging, left/right buttons attached to each thumb.

### Browse pools by thumbnails (of the first, or an assigned "cover" post)
Pools page but with thumbnails in a grid format. The first post in the pool can be the assigned cover. Also include the name of the pool in the grid member.

### Mass pooling like mass tagging
>I'm not sure if it's best to add it as a separate mass-tagging toggle like the one for tags already, or if there should be a "metatag" system like Danbooru 2 where posts tagged like `pool:1234` automatically get assigned to `pool 1234`. That way the mass tagging feature already implemented can also be used to add posts to pools easily.

### Post shows pool info; Browsing a pool from /post page -> See #354 
- However, because a post can belong to multiple pools, the user must select the pool to browse it. This can be done using the pool list. When the user selects one, the main pool bar is displayed and starts browsing the pool pages. Or, pool browsing is only started from a `pool:id` query.
- Pool paging by passing the query, eg. `/post/99/query=pool:7` and selecting the next post relative to the current post ID.  Assuming `pool:id` returns the sorted results by default.

### Whitespace for pool names
> Currently it uses the same validation as tags, and the names are split on whitespace.

PR #403, opened 2021-05-09
Improve pool navigation/browsing

Ruin0x11:
Related to #380.

1. Implements `sort:pool` for post searches, which will sort the results by the order defined by the `pool:` named token in the same search query.
2. Modifies the previous/next post order for the post view if a `pool:` named token is in the current query.
3. Adds a grid-based view for the pools list page, showing up to the first three posts in each pool.

4. Implements pool navigators. They allow navigating to the first/last/next/previous post in each pool the post belongs to. The current pool active in the search query will get bolded/placed at the top.


server/szurubooru/func/posts.py
```diff
@@ -967,3 +972,81 @@ def search_by_image(image_content: bytes) -> List[Tuple[float, model.Post]]:
...
+    around = dict()
+    pool_ids = set()
+    post_ids = set()
+
+    dbquery = """
+    SELECT around.ord, around.pool_id, around.post_id, around.delta
+    FROM pool_post pp,
+        LATERAL get_pool_posts_around(pp.pool_id, pp.post_id) around
+    WHERE pp.post_id = :post_id;
+    """
```
sgsunder, line 989: Use SQLAlchemy instead of writing raw SQL.

server/szurubooru/func/posts.py
```diff
@@ -134,14 +135,18 @@ def get_post_content_path(post: model.Post) -> str:
     )
 
 
-def get_post_thumbnail_path(post: model.Post) -> str:
-    assert post
+def get_post_thumbnail_path_from_id(post_id: int) -> str:
```
sgsunder, line 138: Not sure if this function is necessary, see below

server/szurubooru/func/posts.py
sgsunder: The `search_by_image()` function uses raw SQL only because Postgres' `unnest()` function is not implemented in SQLAlchemy. But this query should be able to be written using SQLAlchemy

sgsunder: You may want to also look at how we currently implement `GET /post/<ID>/around` as a springboard

server/szurubooru/func/posts.py
```diff
@@ -967,3 +972,81 @@ def search_by_image(image_content: bytes) -> List[Tuple[float, model.Post]]:
...
+    post_ids = set()
+
+    dbquery = """
+    SELECT around.ord, around.pool_id, around.post_id, around.delta
+    FROM pool_post pp,
+        LATERAL get_pool_posts_around(pp.pool_id, pp.post_id) around
+    WHERE pp.post_id = :post_id;
+    """
+
+    for order, pool_id, post_id, delta in db.session.execute(dbquery, {"post_id": post.post_id}):
```
sgsunder, line 991: When using SQLAlchemy, you should just capture the `model.Pool` and `model.Post` objects directly instead of working with the IDs

server/szurubooru/func/posts.py
```diff
@@ -967,3 +972,81 @@ def search_by_image(image_content: bytes) -> List[Tuple[float, model.Post]]:
...
+        elif delta == 2:
+            around[pool_id][3] = post_id
+        pool_ids.add(pool_id)
+        post_ids.add(post_id)
+
+    pools = dict()
+    posts = dict()
+
+    for pool in db.session.query(model.Pool).filter(model.Pool.pool_id.in_(pool_ids)).all():
+        pools[pool.pool_id] = pool
```
sgsunder, line 1009: This would be unnecessary if you let SQLAlchemy fetch the pools directly in the first place, instead of fetching the pool IDs

server/szurubooru/func/posts.py
```diff
@@ -967,3 +972,81 @@ def search_by_image(image_content: bytes) -> List[Tuple[float, model.Post]]:
...
+
+    for result in db.session.query(model.Post.post_id).filter(model.Post.post_id.in_(post_ids)).all():
+        post_id = result[0]
+        posts[post_id] = { "id": post_id, "thumbnailUrl": get_post_thumbnail_path_from_id(post_id) }
```
sgsunder, line 1013: This would be unnecessary if you let SQLAlchemy fetch the posts directly in the first place, instead of fetching the post IDs.

Also, when you do this, the `get_post_thumbnail_path_from_id()` won't be needed anymore

server/szurubooru/search/configs/post_search_config.py
```diff
@@ -114,17 +114,55 @@ def _pool_filter(
     query: SaQuery, criterion: Optional[criteria.BaseCriterion], negated: bool
 ) -> SaQuery:
     assert criterion
-    return search_util.create_subquery_filter(
-        model.Post.post_id,
-        model.PoolPost.post_id,
-        model.PoolPost.pool_id,
-        search_util.create_num_filter,
-    )(query, criterion, negated)
+    from szurubooru.search.configs import util as search_util
```
sgsunder, line 117: Already imported at top

server/szurubooru/migrations/functions.py
```diff
@@ -0,0 +1,68 @@
+from alembic_utils.pg_function import PGFunction
```
sgsunder, line 1: Is writing a SQL function really necessary instead of doing it in SQLAlchemy? This is something that can get very hard to maintain in the future

sgsunder (changes requested):

sgsunder:
Not sure if this is still active but I've added some comments

parallax4:
Are there any updates on this? Thanks.

acctfpn:
@neobooru @sgsunder
Would this be mergeable if the comments get addressed? I may look into it if that's the case.

Issue #416, opened 2021-06-28
OpenGraph embeds


ioistired:
It would be nice if sharing a link to a szurubooru post on supported platforms (Twitter, Facebook, chat clients etc) would embed a preview of the post. However, as far as I'm aware (not sure though) this requires the metadata to be embedded in the HTML, rather than added by client side javascript, so it might require the backend to also start serving frontend pages or something along those lines.

robobenklein:
Here's my proposed design for enabling OpenGraph tags without changing the structure of the single-page app (SPA), based on the idea from https://stackoverflow.com/questions/50263513/redirect-crawlers-to-internal-microservice-in-nginx

1. Request comes in from crawler / bot
2. Nginx performs a match against the user agent (if it's not a crawler/bot, serve normally)
3. For bots, redirect internally (as the reverse proxy) to `http://backend/opengraph/$1`
4. Server backend generates an HTML stub page with just the opengraph tags (this is where permissions will be handled - e.g. denying access if the default post access level is not public)

Step 2 is the part where I am not sure it's the best implementation, since matching the user agent is an evolving problem which needs to be updated for every service that OG tags would be displayed on. (For example the discordbot user agent is not handled in the stackoverflow post I linked: `"Mozilla/5.0 (compatible; Discordbot/2.0; +https://discordapp.com)"`)

ioistired:
Or we could make the frontend a server side app which serves static HTML except for all pages that have OpenGraph tags, for which it serves static HTML + the tags

robobenklein:
@ioistired how would the frontend go about retrieving that content for the og tags from the backend server? AFAIK the current nginx config has no serverside processing going on at all right now, so what would that additional code look like?

ioistired:
There'd be a backend server for the frontend. index.htm currently looks like this:

```html
<!DOCTYPE html>
<html>
 <head>
  <meta charset="utf-8"/>
  <meta content="width=device-width,initial-scale=1,maximum-scale=1" name="viewport"/>
  <meta content="#24aadd" name="theme-color"/>
  <meta content="yes" name="apple-mobile-web-app-capable"/>
  <meta content="black" name="apple-mobile-web-app-status-bar-style"/>
  <meta content="#ffffff" name="msapplication-TileColor"/>
  <meta content="/img/mstile-150x150.png" name="msapplication-TileImage"/>
  <title>
   Loading...
  </title>
  <!-- snip -->
 </head>
 <body>
  <!-- snip -->
[...]

PR #422, opened 2021-07-10
Nginx / Lua based approach to Open Graph tags

robobenklein:
This is an alternative option for #416 which I believe has the following benefits compared to #421 

- Does not require any additional processes, containers, or listening ports
- Much lower memory & cpu footprint vs an additional python server ( =handling more traffic)
- Smaller container size vs adding another Python webserver container

There are some issues with this implementation which I want to fix first though:

- The openresty image is much larger than the bare nginx image (101MB vs 22MB) so I would like to just add on the Lua module to the Nginx image ourselves (22+??MB) (although there are some additional complications with that process which I am still sorting out, like runtime Lua optimizations being missing in the manual install on nginx)
-

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -18,6 +18,7 @@ COPY docker-start.sh /
 
 WORKDIR /etc/nginx
 COPY nginx.conf.docker ./nginx.conf
+COPY metatags.lua .
 
 WORKDIR /var/www
 COPY --from=builder /opt/app/public/ .
@@ -26,7 +27,8 @@ COPY --from=builder /opt/app/public/ .
 FROM openresty/openresty:alpine-fat as release
 
 RUN apk --no-cache add dumb-init
-RUN /usr/local/openresty/luajit/bin/luarocks install gumbo
+RUN opm get bungle/lua-resty-template
+# RUN /usr/local/openresty/luajit/bin/luarocks install gumbo
 COPY --from=approot / /
 
 CMD ["/docker-start.sh"]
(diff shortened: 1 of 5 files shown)


robobenklein:
A tangentially related idea that I had while writing this, since the lua is embedded within Nginx it's possible to return the correct logical HTTP status codes for resources which might not exist.

e.g. making an initial request to `/user/does-not-exist` could return a 404 status while still serving the single page app, and would allow external sites to better understand the real status of a linked page.

sgsunder:
Can't we just add a route to our existing REST application that will generate these stub HTML files instead? It wouldn't require adding a dependency to Lua and it would be more lightweight.

robobenklein:
@sgsunder absolutely possible, I just didn't take that approach because it involved the backend containing some of the code from the frontend

I can try that approach, just need to address the following points:

- How should I pass the html SPA template to the backend? (Or should it just take ownership of that file entirely?)
- How should the backend handle templating the html file? (AFAIK it currently only returns json data)

Then I think I could try that approach and compare. Right now before I even have a plan I think it'll turn into a [faster response and larger frontend container] vs [slim frontend and slightly higher response times].

Another point to research is whether the output from the python API can be sent chunked, so that the browser can start loading the SPA assets before all the page tag information has been fetched by the server. (You can see where I do this in the lua by flushing the output.)

sgsunder:
@robobenklein I made PR #441 trying this out, I'm open to any feedback and we can discuss which implementation would be more efficient.

Issue #415, opened 2021-06-10
Broken markdown ATX-style headings and broken URL links


dankles:
Greetings Team,

I've found two bugs with parsing **Comment** or **Tag Description** textareas. If I need to create separate Issues for these, let me know.

## Issue 1 - Broken Headings
If I use an ATX style heading on the first line of a **Comment** or **Tag Description** textarea, it doesn't format the heading. 

**Example 1**: This will stay as `# H1`, rather than format correctly, if placed on the first line:
```
# H1
```
`## H2` and `### H3` also won't format when used on the first line.

**Example 2**: If I lead the textarea with a newline, it formats fine:
```

# H1
```

**Example 3**: Underline heading style works fine on the first line:
```
H1
===
```
This works too:
```
H2
---
```

## Issue 2 - Broken URLs
A URL like this `https://example.com/user/bob/profile`, will link to  this `https://example.com+bob/profile`, rather than the correct link. `/user/<string>` seems to be getting parsed weirdly.

If I encapsulate a URL like that in a markdown link, it translates badly as well.
**Example**: `[Description](https://example.com/user/bob/profile)` will still link to `https://example.com+bob/profile`.

neobooru:
Thanks for reporting these issues. The note handling code is quite complex, so I don't really expect these issues to get fixed very soon.

A possible workaround for now:
- You can link to an user by using `+username`
- You can link to an post by using `@postnumber`, e.g. `@12`
- You can link to a tag by using `#tag_name`

This all also works inside link tags, e.g. `[tag](#tag_name)`

po5:
Issue 1 is caused by overzealous escaping of #'s https://github.com/po5/szurubooru/commit/d49f179c68e0a1dd3486d2e73e744eb072829059

PR #429, opened 2021-08-08
Fixed markup formatting to only parse host urls

realwumbl3:
Related to #415. (Issue 2 - Broken URLs)

  * Client code was parsing /user/, /post/ and /tag/ urls from every site. Now only parses internal links.

### DIFF
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -65,16 +65,11 @@ class TagPermalinkFixWrapper extends BaseMarkdownWrapper {
 // post, user and tags permalinks
 class EntityPermalinkWrapper extends BaseMarkdownWrapper {
     preprocess(text) {
+        const re = (template) => new RegExp(`\\b(https?:\/\/(?:www\.|(?!www))${window.location.host}|www\.${window.location.host}|https?:\/\/(?:www\.|(?!www))${window.location.host}|www\.${window.location.host})${template}\\b`, "g")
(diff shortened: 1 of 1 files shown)


sgsunder:
> Though honestly I personally feel like the markdown util shouldn't touch any links at all. Just use +user and #tag, and leave our links as is. That said, I don't have the authority to make such a breaking decision.

I agree with this, I don't think the parser should try to 'fix' links like that

neobooru:
> I agree with this, I don't think the parser should try to 'fix' links like that

Should we just remove it then? It would be a breaking change, though I doubt that many actually used that feature.

sgsunder:
I think we should. I noticed this very bug before, when I used internal links within markdown syntax:
```
[Really cool art!](https://booru.example.com/post/1337)
```

would end up getting mangled to:
```
[Really cool art!](https://booru.example.com@1337)
```

I don't know if there's an easy way of fixing this and OP's problem without adding a bunch of parsing logic, and I agree that I doubt many people would notice or care

Telkhine:
The URL rewriting code introduced several bugs into the application. I'd prefer it to be removed, if the issues can't easily be resolved otherwise.

neobooru:
Seeing as no one objected to removing the URL handling code I went ahead and merged #444. If someone suddenly starts missing this feature then please feel free to open a new issue.

PR #441, opened 2021-09-24
Implement server-generated Open Graph tags

sgsunder:
@robobenklein this was my quick attempt at Open Graph. There's still a lot more tag data that needs to be generated but you should get the idea. 

Implements #416 , Closes #421 and #422

### DIFF
--- a/client/nginx.conf.docker
+++ b/client/nginx.conf.docker
@@ -69,9 +69,8 @@ http {
             error_page 404 @notfound;
         }
 
-        location / {
+        location ~ ^/(js|css|img|fonts)/.*$ {
             root /var/www;
-            try_files $uri /index.htm;
 
             sendfile on;
             tcp_nopush on;
@@ -79,6 +78,16 @@ http {
 
             gzip_static on;
             gzip_proxied expired no-cache no-store private auth;
+
+            error_page 404 @notfound;
+        }
+
+        location / {
+            tcp_nodelay on;
+
+            proxy_pass http://backend;
+
+            error_page 500 502 503 504 @badproxy;
         }
 
         location @unauthorized {
--- a/server/szurubooru/api/info_api.py
+++ b/server/szurubooru/api/info_api.py
@@ -64,3 +64,31 @@ def get_info(ctx: rest.Context, _params: Dict[str, str] = {}) -> rest.Response:
         )
         ret["featuringTime"] = post_feature.time if post_feature else None
     return ret
+
+
+@rest.routes.get(r"/manifest\.json")
+def generate_manifest(
+    ctx: rest.Context, _params: Dict[str, str] = {}
+) -> rest.Response:
+    return {
+        "name": config.config["name"],
+        "icons": [
+            {
+                # TODO: Host Header and Proxy Prefix
+                "src": "img/android-chrome-192x192.png",
+                "type": "image/png",
+                "sizes": "192x192",
+            },
+            {
+                # TODO: Host Header and Proxy Prefix
+                "src": "img/android-chrome-512x512.png",
+                "type": "image/png",
+                "sizes": "512x512",
+            },
+        ],
+        # TODO: Host Header and Proxy Prefix
+        "start_url": "/",
+        "theme_color": "#24aadd",
+        "background_color": "#ffffff",
+        "display": "standalone",
+    }
--- a/server/szurubooru/rest/app.py
+++ b/server/szurubooru/rest/app.py
@@ -11,15 +11,19 @@ from szurubooru.rest import context, errors, middleware, routes
 
 
 def _json_serializer(obj: Any) -> str:
-    """ JSON serializer for objects not serializable by default JSON code """
+    """JSON serializer for objects not serializable by default JSON code"""
     if isinstance(obj, datetime):
         serial = obj.isoformat("T") + "Z"
         return serial
     raise TypeError("Type not serializable")
 
 
-def _dump_json(obj: Any) -> str:
-    return json.dumps(obj, default=_json_serializer, indent=2)
+def _serialize_response_body(obj: Any, accept: str) -> str:
+    if accept == "application/json":
+        return json.dumps(obj, default=_json_serializer, indent=2)
+    if accept == "text/html":
+        return obj.getvalue()
+    raise ValueError("Unhandled response type %s" % accept)
 
 
 def _get_headers(env: Dict[str, Any]) -> Dict[str, str]:
@@ -72,14 +76,21 @@ def _create_context(env: Dict[str, Any]) -> context.Context:
 def application(
     env: Dict[str, Any], start_response: Callable[[str, Any], Any]
 ) -> Tuple[bytes]:
+    accept = None
     try:
         ctx = _create_context(env)
-        if "application/json" not in ctx.get_header("Accept"):
+        if "application/json" in ctx.get_header("Accept"):
+            accept = "application/json"
+        elif "text/html" in (ctx.get_header("Accept") or "text/html"):
+            accept = "text/html"
+        else:
             raise errors.HttpNotAcceptable(
-                "ValidationError", "This API only supports JSON responses."
+                "ValidationError",
+                "This API only supports the following response types: "
+                ", ".join(routes.routes.keys()),
             )
 
-        for url, allowed_methods in routes.routes.items():
+        for url, allowed_methods in routes.routes[accept].items():
             match = re.fullmatch(url, ctx.url)
             if match:
                 if ctx.method not in allowed_methods:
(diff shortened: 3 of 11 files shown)


sgsunder:
Todo list:

- [ ] Complete the set of OpenGraph tags
- [ ] Generate content URLs more intelligently using `X-Forwarded-Host` header, `X-Forwarded-Prefix` header, and `config.config['domain']`
- [ ] Do we need to add some security features to nginx configuration for proxying the html requests?
- [ ] Write unit tests for REST api changes.

robobenklein:
Ah, this has a lot of useful side effects like being able to change the app icon (maybe we add a setting for it) and potentially return 404's correctly for pages that don't exist.

The only thing left that I'm not sure about is if we can start the response to the client before we have to wait to retrieve post information from the db, which I think is worthwhile to help users on slower connections or with underpowered servers load the SPA resources. Once that's solved then I'd say there's no reason left at all to use the lua approach.

sgsunder:
I think we can try to configure NGINX to ensure it caches the generated html data, which would help keep server-side load down. I'm not sure if we have to worry about slow clients, though, as generating html should only add a bit of latency but not really change the bandwidth demand. (The overwhelming majority of bandwidth demand would still be from static content either way).

Adding configurable app images should probably be done as a separate PR after this one is merged.

robobenklein:
Indeed, setting nginx to cache only the html responses from the server makes sense.

In terms of slow clients I was talking more about TTFB rather than total load time, but if nginx serves the html from cache without a need to hit the DB then there's no issue at all there.

We'll have to gather some data about the best cache settings for people, but perhaps we can just set things like the cache timeout and max size via the env?

sgsunder:
I think using these directives would work for that issue:
> The updating parameter to the proxy_cache_use_stale directive, combined with enabling the proxy_cache_background_update directive, instructs NGINX to deliver stale content when clients request an item that is expired or is in the process of being updated from the origin server. All updates will be done in the background. The stale file is returned for all requests until the updated file is fully downloaded.

([link](https://www.nginx.com/blog/nginx-caching-guide/#Fine%E2%80%91Tuning-the-Cache-and-Improving-Performance))

client/nginx.conf.docker
```diff
@@ -19,6 +19,9 @@ http {
     server_tokens off;
     keepalive_timeout 65;
 
+    proxy_cache_path /tmp/nginx-cache
+        levels=1:2 keys_zone=spa_cache:1m max_size=50m inactive=60m use_temp_path=off;
```
robobenklein, line 23: This is the part here where I wanted to try to set these values via environment variables.

No clue if it's easy or not, but I'll at least try to look around.

sgsunder: We could just do a sed substitution like we do for the backend hostname.

server/szurubooru/rest/app.py
```diff
@@ -37,6 +41,11 @@ def _create_context(env: Dict[str, Any]) -> context.Context:
     path = path.encode("latin-1").decode("utf-8")  # PEP-3333
     headers = _get_headers(env)
 
+    if config.config["domain"]:
+        url_prefix = config.config["domain"].rstrip("/")
+    else:
+        url_prefix = headers.get("X-Forwarded-Prefix", "")
```
robobenklein, line 47: I feel like this could end up being potentially dangerous for anyone who doesn't explicitly set this variable, since it essentially lets the user put whatever they want here. I can't immediately think of how I would exploit it for major damage, but it's been a component in other bigger security problems.

https://github.com/traefik/traefik/security/advisories/GHSA-6qq8-5wq3-86rp

I personally think the server needs to be in control of what values could be accepted here in all cases.

robobenklein:
I think I figured out how I would exploit `X-Forwarded-Prefix`,

- Completely destroy the utility of the cache and force disk thrashing by changing the prefix in repeated attempts with new values
- Poison the cache to serve incorrect responses by setting the prefix to a path which already exists for a normal page
- If the reverse proxy doesn't protect the `.well-known` directory (or similar security convention like robots.txt), then there's some *serious* damage potential by poisoning the cache there with bad data (like I could completely deny the ability for an automated HTTPS certificate renewal)

So basically, I *strongly* recommend against using the `X-Forwarded-Prefix` to generate URLs. I think allowing only a singular, config-defined prefix is the secure method.

sgsunder:
Hmm, OK, I'll make a commit swapping it over to using a config file only. Shouldn't be too hard

client/nginx.conf.docker
sgsunder: Since we're only caching the HTML stub pages, and each stub page is at most 3.5 kB, the 50 MB cache size would be able to store about 14,000 stub pages for about 60 minutes each. I don't really see this happening for the majority of users, and those that would see this will probably fork the project anyway to make more involved source code modifications.

server/szurubooru/api/info_api.py
```diff
@@ -64,3 +64,28 @@ def get_info(ctx: rest.Context, _params: Dict[str, str] = {}) -> rest.Response:
         )
         ret["featuringTime"] = post_feature.time if post_feature else None
     return ret
+
+
+@rest.routes.get(r"/manifest\.json")
```
robobenklein: This might be entirely a non-issue, but is it possible that a request would come to this endpoint with `Accept: application/json`? (because I see that all the accept:json requests not to /api get denied in the nginx conf)

sgsunder: I've redefined the location of the manifest file to be `/api/manifest.json` instead of `/manifest.json` (see [here](https://github.com/rr-/szurubooru/blob/17152f68a4b8594c8c72874936e2ed7028846d10/server/szurubooru/api/opengraph_api.py#L80)) to accommodate this.

robobenklein:
Found one more problem to address: Discordbot/2.0 does not send an `Accept` header at all, of any kind...

```
GET /post/4582 HTTP/1.0
Host: ----
Connection: upgrade
X-Forwarded-Prefix: /
X-Forwarded-For: ----
X-Scheme: https
X-Real-IP: ----
X-Forwarded-Proto: https
user-agent: Mozilla/5.0 (compatible; Discordbot/2.0; +https://discordapp.com)


>>> HTTP/1.1 406 Not Acceptable
Server: nginx
Date: Tue, 22 Mar 2022 18:31:05 GMT
Content-Type: application/json
Content-Length: 126
Connection: close

{
  "name": "ValidationError",
  "title": "Not Acceptable",
  "description": "This route only supports text/html responses."
}
```

Can be replicated without sticking links into discord every test by unsetting the header with httpie:

```http
# http -p Hh https://your-boorus.cool-address/post/4582 Accept:
GET /post/4582 HTTP/1.1
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: ----
User-Agent: HTTPie/2.4.0

HTTP/1.1 502 Bad Gateway
Connection: keep-alive
Content-Length: 40
Content-Type: text/plain
Date: Tue, 22 Mar 2022 18:40:33 GMT
Server: nginx
```

Without a client Accept: header, we need the server to still return content in order for embeds to work on sites like Discord.

server/config.yaml.dist
```diff
@@ -3,11 +3,28 @@
...
-# full url to the homepage of this szurubooru site, with no trailing slash
-domain: # example: http://example.com
 # used to salt the users' password hashes and generate filenames for static content
 secret: change
 
+# set to the root web address for your instance
+# example values:
+# - `/` (default) is used when the domain is unknown
+# - `https://szuru.example.com/` if you know the specific domain
+#      and is required if you want email-based password reset
```
robobenklein, line 13: Also required for Twitter cards to work, as they require absolute (not relative) content URLs in the meta tags. (https://twittercommunity.com/t/card-error-unable-to-render-or-no-image-read-this-first/62736)

server/szurubooru/func/posts.py
```diff
@@ -264,7 +251,7 @@ def serialize_relations(self) -> Any:
             {
                 post["id"]: post
                 for post in [
-                    serialize_micro_post(rel, self.auth_user)
+                    serialize_micro_post(rel, self.auth_user, self.url_prefix)
```
robobenklein, line 254:
```
server_1  |   File "/opt/app/szurubooru/func/posts.py", line 254, in <listcomp>
server_1  |     serialize_micro_post(rel, self.auth_user, self.url_prefix)
server_1  | AttributeError: 'PostSerializer' object has no attribute 'url_prefix'
```

Also it doesn't appear to take that third argument at all?

robobenklein:
Fix for discord not using any `Accept` header:

```diff
diff --git a/client/nginx.conf.docker b/client/nginx.conf.docker
index 1ce531d7..b9e3dba4 100644
--- a/client/nginx.conf.docker
+++ b/client/nginx.conf.docker
@@ -92,6 +92,12 @@ http {
         location / {
             tcp_nodelay on;
 
+            set $client_accept "text/html";
+            if ($http_accept) {
+                set $client_accept $http_accept;
+            }
+            proxy_set_header Accept $client_accept;
+
             # remove unneeded auth headers to improve caching
             proxy_set_header Authorization "";
```

PR #442, opened 2021-10-01
add support to flv by flv.js

uonr:
[flv.js](https://github.com/Bilibili/flv.js/)

now szurubooru can download and play flv video.

### DIFF
--- a/client/js/views/post_main_view.js
+++ b/client/js/views/post_main_view.js
@@ -50,6 +50,20 @@ class PostMainView {
             postContainerNode.querySelector(".post-overlay"),
             ctx.post
         );
+        // suppurt flv play
+        if (ctx.post.mimeType === "video/x-flv") {
+            const flvjs = require("flv.js");
+            if (flvjs.isSupported()) {
+                const video = document.querySelector(".post-content video");
+                const src = video.querySelector("source").src
+                const player = flvjs.createPlayer({
+                    type: 'flv',
+                    url: src,
+                });
+                player.attachMediaElement(video);
+                player.load();
+            }
+        }
 
         if (ctx.post.type === "video" || ctx.post.type === "flash") {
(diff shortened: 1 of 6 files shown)


client/js/views/post_main_view.js
```diff
@@ -50,6 +50,20 @@ class PostMainView {
             postContainerNode.querySelector(".post-overlay"),
             ctx.post
         );
+        // suppurt flv play
```
sgsunder, line 53: Don't really need this comment

client/js/views/post_main_view.js
```diff
@@ -50,6 +50,20 @@ class PostMainView {
...
+        if (ctx.post.mimeType === "video/x-flv") {
+            const flvjs = require("flv.js");
```
sgsunder, line 55: Any reason why this isn't at the top of the file alongside the other requires?

server/szurubooru/func/mime.py
```diff
@@ -55,14 +58,17 @@ def get_extension(mime_type: str) -> Optional[str]:
...
         "application/octet-stream": "dat",
     }
     return extension_map.get((mime_type or "").strip().lower(), None)
 
-
 def is_flash(mime_type: str) -> bool:
     return mime_type.lower() == "application/x-shockwave-flash"
 
+def is_flv(mime_type: str) -> bool:
+    return mime_type.lower() == "video/x-flv"
```
sgsunder, line 70: I think it would be better if we added this mime type to the `is_video()` function instead of creating a new function.

sgsunder (changes requested):
Can you include the new mime type to the display list in `post_readonly_sidebar.tpl`?

You should also add the new mime type `video/x-flv` to the following unit tests:
- `test_update_post_content_for_new_post`
- `test_is_video`
- `test_get_mime_type`

PR #445, opened 2021-10-07
server: post category filter

### LINKED ISSUE #438: Filter posts by containing tag with category?
Is it possible to filter posts by whether or not it contains any tags within a category?
For example, I have a artist category and I want to filter out images that have any tag belonging to that category, so that the remaining images are those which don't have an artist tag.

If not, would it be possible to implement?
I imagine the simple way is to get a table containing all artist tags then filter all of those tags against posts. Then, a new token like `tagged-with-category` or `has-tags-from-category`
I probably don't know what I'm talking about, I only took an SQL class.

neobooru:
Work in progress, my SQL is a bit rusty.

Fixes #438

### DIFF
--- a/server/szurubooru/search/configs/post_search_config.py
+++ b/server/szurubooru/search/configs/post_search_config.py
@@ -122,6 +122,34 @@ def _pool_filter(
     )(query, criterion, negated)
 
 
+def _category_filter(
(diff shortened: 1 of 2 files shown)


neobooru:
The code isn't pretty, but it works fine I guess.

PR #450, opened 2021-11-28
updated build files

skybldev:
- `npm i -g npm@lts` (in client/Dockerfile) is no longer valid as per https://github.com/npm/cli/wiki/Support-Policy#long-term-support-lts
- updated pre-commit config to use latest repos

the commit message somehow omitted `npm i -g npm@lts`, weird


sgsunder:
Merged, thanks!

Issue #456, opened 2022-01-05
Feature Request: Mass deletion of posts 


FlipperLP:
Would it be possible to implement a mass deletion-feature? 

I recently made an automation to add posts from my submission box on furaffinity. But sometimes it's necessary to clean up some posts, so wanted to ask if there is any chance to add a mass deletion-feature. It would function the same way as Mass Tag or Mass edit safety.

neobooru:
I started working on the frontend for this. How does this look? The "Delete x posts" button looks a bit out of place imo, but I wouldn't know where else to put it.

![img](https://user-images.githubusercontent.com/50623835/149601372-bddfd3a9-6d56-4a82-b3cb-afde4e05e057.png)
![img](https://user-images.githubusercontent.com/50623835/149601379-56e8a075-3c2a-42e9-adfc-59545f1cc90d.png)

FlipperLP:
I think the other 2 buttons feel out of place as well, maybe they should be moved in the future, but for now I think it's fine to keep them where they are.
Thank you for visualizing it!

pink-red:
> The "Delete x posts" button looks a bit out of place imo

Maybe moving everything but "Syntax help" to the right side of the page would help? So that there's everything search-related on the left and bulk actions on the right.

But overall I would agree with @FlipperLP that this is already fine.

pink-red:
Something like this:
![image](https://user-images.githubusercontent.com/94079189/149638741-f39aa66c-3fe7-44ff-836e-240748d95561.png)
![image](https://user-images.githubusercontent.com/94079189/149638776-0609593c-13ef-48d1-869d-c785954269b3.png)

studiomav:
Any such solution would be greatly appreciated...

panda-lsy:
Any such solution would be greatly appreciated❤

neobooru:
#459 has been merged, so I'll close this issue. Please create a new issue if something is broken or can be improved.

PR #459, opened 2022-01-15
client: add bulk delete feature

neobooru:
Implement feature request #456

### DIFF
--- a/client/css/post-list-view.styl
+++ b/client/css/post-list-view.styl
@@ -114,6 +114,29 @@
                     &[data-disabled]
                         background: rgba(200, 200, 200, 0.7)
 
+                .delete-flipper
+                    display: inline-block
+                    padding: 0.5em
+                    box-sizing: border-box
+                    border: 0
+                    &:after
+                        display: inline-block
+                        width: 1em
+                        height: 1em
+                        text-align: center
+                        line-height: 1em
+                        font-size: 2.2em
+                    &.delete
+                        background: rgba(255, 0, 0, 0.7)
+                        &:after
+                            color: white
+                            font-family: FontAwesome;
+                            content: "\f1f8"; // fa-trash
+                    &:not(.delete)
+                        background: rgba(200, 200, 200, 0.7)
+                        &:after
+                            color: white
+                            content: '-'
 
             .thumbnail
                 width: 100%
@@ -215,7 +238,19 @@
         .append
             @media (max-width: 1000px)
                 margin-left: 0
-
+    .bulk-edit-delete
+        &.opened
+            .start
+                @media (max-width: 1000px)
+                    margin-left: 0
+        &:not(.opened)
+            .start
+                display: none
+        .append.open
+            @media (max-width: 1000px)
+                margin-left: 0
+        .start
+            margin-left: 1em
     .safety
         margin-right: 0.25em
         &.safety-safe
--- a/client/html/posts_header.tpl
+++ b/client/html/posts_header.tpl
@@ -28,4 +28,11 @@
             %><a href class='mousetrap button append close'>Stop editing safety</a><%
         %></form><%
     %><% } %><%
+    %><% if (ctx.canBulkDelete) { %><%
+        %><form class='horizontal bulk-edit bulk-edit-delete'><%
+            %><a href class='mousetrap button append open'>Mass delete</a><%
+            %><input class='mousetrap start' type='submit' value='Delete selected posts'/><%
+            %><a href class='mousetrap button append close'>Stop deleting</a><%
+        %></form><%
+    %><% } %><%
 %></div>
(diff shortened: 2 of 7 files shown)


client/css/post-list-view.styl
```diff
@@ -114,6 +114,29 @@
...
+                        background: rgba(255, 0, 0, 0.7)
+                        &:after
+                            color: white
+                            font-family: FontAwesome;
+                            content: "\f1f8"; // fa-trash
+                    &:not(.delete)
+                        background: rgba(200, 200, 200, 0.7)
+                        &:after
+                            color: white
+                            content: '-'
```
neobooru, line 139: Not sure which symbol to use here.

client/html/posts_header.tpl
```diff
@@ -28,4 +28,11 @@
             %><a href class='mousetrap button append close'>Stop editing safety</a><%
         %></form><%
     %><% } %><%
+    %><% if (ctx.canBulkDelete) { %><%
+        %><form class='horizontal bulk-edit bulk-edit-delete'><%
+            %><a href class='mousetrap button append open'>Mass delete</a><%
+            %><input class='mousetrap start' type='submit' value='Delete selected posts'/><%
```
neobooru, line 34: It would be cool to show the text "Delete {Count} posts", but I couldn't think of a good implementation for that.

client/js/controllers/post_list_controller.js
```diff
@@ -91,6 +103,36 @@ class PostListController {
...
+    }
+
+    async _evtDeleteSelectedPosts(e) {
+        if (
+            confirm(
+                `Are you sure you want to delete ${this._bulkEditDelete.length} posts?`
+            )
+        ) {
+            for (let postId of this._bulkEditDelete) {
+                const post = await Post.get(postId);
```
neobooru, line 126: There is probably a better way to get a reference to the post.

client/js/controllers/post_list_controller.js
```diff
@@ -44,13 +45,24 @@ class PostListController {
...
+            canBulkDelete: api.hasPrivilege("posts:bulk-edit:delete"),
             bulkEdit: {
                 tags: this._bulkEditTags,
             },
         });
         this._headerView.addEventListener("navigate", (e) =>
             this._evtNavigate(e)
         );
+        // This doesn't feel like the best solution
+        this._headerView._bulkDeleteEditor.addEventListener(
```
neobooru, line 57: This isn't pretty, because we call addEventListener on a (theoretically speaking) private field.

client/js/controllers/post_list_controller.js
neobooru: Oh, and this (currently) also doesn't show the progress nor does it catch the errors in a nice way.

client/js/controllers/post_list_controller.js
```diff
@@ -44,13 +45,24 @@ class PostListController {
...
+            "deleteSelectedPosts",
+            (e) => {
+                this._evtDeleteSelectedPosts(e);
+            }
+        );
+
+        // Contains the id of each post that we want to delete (while using the bulk delete feature).
+        this._bulkEditDelete = [];
```
neobooru, line 65: We could also just use `document.querySelectorAll("... > .delete")` to manage our state, but that feels dirty.

parallax4:
@neobooru What's the status on this? Does it currently have basic functionality, or is it missing some important pieces?

neobooru:
The current implementation looks like this:

firefox_2023_01_11_00_18_1080
firefox_2023_01_11_00_18_1081
firefox_2023_01_11_00_18_1082

I haven't moved the buttons to the right [as suggested here](https://github.com/rr-/szurubooru/issues/456#issuecomment-1013759304). That'll probably be a separate PR as the mass tagger also isn't 100% correct on all resolutions, and fixing that would bloat this PR.

I think this feature is pretty much production ready now. Sure it could be a bit nicer (progress bar etc), but that takes more energy than I'm currently willing to invest. If anyone sees any issues, or has suggestions, please share them. If nothing pops up then I'll probably merge this by the end of this week(tm).

recordcrash:
This seems to cause an error on mobile, it's not production ready.

"this._headerView._bulkDeleteEditor.addEventListener" leads to Uncaught TypeError: Cannot read properties of undefined (reading 'addEventListener') on mobile only (maybe the mass delete button doesn't appear there?), which was a pain to track down. 

This error leads to pages not loading unless people interact with the interface manually after loading. For now I'm going to aim at a previous commit or something.

negiworx:
As a comment on this I was just using the mass delete on my cellphone (Note 5g Ultra) and there were no issues with it and I had done the mass delete multiple times. so the issue may only appear on some browsers and or OS.

PR #464, opened 2022-02-12
Make waitress thread count configurable.

### LINKED ISSUE #463: Performance degrades after half a million posts
The board was very responsive up to 300-400K posts and now it takes ages to load. Is there any way to increase the Worker Count? the entire board seems to be synchronous and locking up. I can see that the waitress is very slow at processing everything.

Especially when i want to apply tags to many images.

Sometimes i get a

```
Request has been terminated
Possible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin, the page is being unloaded, etc.
```

when i want to use the rest api for adding tags, 
`'Failed to connect to szurubooru REST API'` appears after some time.

```
22-02-08 21:40:11] waitress.queue Task queue depth is 24
server_1  | [2022-02-08 21:40:11] szurubooru.middleware.request_logger GET /post/399256 (user=shitcord, queries=8)
server_1  | [2022-02-08 21:40:11] waitress.queue Task queue depth is 24
server_1  | [2022-02-08 21:40:12] szurubooru.middleware.request_logger GET /post/522463 (user=shitcord, queries=6)
server_1  | [2022-02-08 21:40:12] waitress.queue Task queue depth is 24
```

noirscape:
This should fix most scaling problems without needing to start
more server instances. By default, waitress maintains at most
4 threads. This works fine if the database is small (sub 100k posts)
but causes a large Task queue depth to occur if the database is larger.

Letting users increase the amount of threads means that one server instance
is able to handle more requests without locking up the rest of the site.

This adds a new environment variable to .env, THREADS, which can be used to
configure the amount of threads to start and is by default set to 4
(the default amount used by waitress).

In addition, this pretty much closes #463.


sgsunder:
Merged in 82541536afd95ebbbb72b1983caae77b67eb8b2d, thanks for the PR!

Issue #382, opened 2021-03-23
Feature Request: Mass Safety/Tag edit on upload


SkandhaR:
I want to start by saying, this tool is awesome!

Like a lot of people I have my files somewhat sorted. So the process of individually selecting each post in the Mass Tag/Safety Edit feels like wasted energy when there are 20+ items being edited at a time.
I see that this function can be achieved through the API. But I'd like to see something like this in the Upload GUI. 
![image](https://user-images.githubusercontent.com/34901480/112224822-19b70e00-8be9-11eb-9707-1c3fff105257.png)
I understand this is basically only an issue when doing bulk uploads through the web. API might be the answer to all problems, but just because I can deploy a docker container doesn't mean I can figure out how to use the API.😅

If this isn't something you're not interested in implementing, can I have some info on the following:
- Is there a way to filter by "untagged" posts?'
or
- Is there a way to set a default tag for uploads?

neobooru:
I can only unfortunately only answer one question, but you can search for untagged posts using `tag-count:0`

noirscape:
Would like to see this too. I often tend to upload CG sets, which are most
commonly distributed as ZIP files. Manually tagging these after upload,
even with the mass tagger (which has its own problems) is a pain.

SkandhaR:
> I can only unfortunately only answer one question, but you can search for untagged posts using `tag-count:0`

Thanks! Apparently I'm blind and didn't look at the "Syntax Help" button...

neobooru:
Another option would be to just enhance the already existing mass tagger. I personally use a custom application for my mass tagging needs which allows me to select multiple posts at once and add tags to them all in one click (aka not clicking the + button 100+ times). Same goes for the safety.

404Fox:
This is probably what you guys are looking for
https://github.com/rr-/szurubooru/commit/035ce19788626c007619c004ef3b702f70ac262e
https://github.com/rr-/szurubooru/commit/902cca66f3062d6c8c93b4bb357d0522d7937d73

404Fox:
I mean just merge the fork whole
or someone could copy/paste into a new commit here, preferably that person being the one who wrote it.

sgsunder:
> I mean just merge the fork whole
or someone could copy/paste into a new commit here, preferably that person being the one who wrote it.

Yup, that's the issue. The fork seems to have been deleted/privated so I can't access any parent commits of it to merge it myself.

dkeatin2:
[...]

PR #468, opened 2022-03-06
Set common tags for all uploads, with auto-completion

noirscape:
This cherrypicks @Hunternif's https://github.com/hunternif/szurubooru/commit/035ce19788626c007619c004ef3b702f70ac262e commit to add common tags to uploads.

Relates to #382, although it does not implement mass safety changing.

Again, almost all credit goes to @Hunternif for the actual implementation.

### DIFF
--- a/client/css/post-upload.styl
+++ b/client/css/post-upload.styl
@@ -15,6 +15,7 @@ $cancel-button-color = tomato
         &.inactive .skip-duplicates
         &.inactive .always-upload-similar
         &.inactive .pause-remain-on-error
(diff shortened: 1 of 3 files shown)


noirscape:
(note: build failure is unclear to me; it seems there is an issue installing server container packages with GitHub actions? regardless, this PR does not touch the server...)

neobooru:
I will close this PR as #534 is an 'evolution'/improved version of this PR. Please take a look at that PR if you have any strong opinions about this feature. Thanks for the picking out the correct commits!

PR #475, opened 2022-03-28
Added BuildKit flags fix to INSTALL.md

skybldev:
Added this because recently, there have been more problems with `docker-compose build` where it errors:

    ERROR: Service 'server' failed to build: failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument

Recent Docker versions have switched to using `buildx` (BuildKit) to build containers, but that needs to be enabled, either in `daemon.json` or through an environment variable. But since we are using Docker Compose, it doesn't pass it to Docker; so the environment variable needs to be set. At least that's what I've heard and figured out sweat_smile My explanation might be very wrong - but it works :)

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -34,33 +34,79 @@ and Docker Compose (version 1.6.0 or greater) already installed.
     Read the comments to guide you. Note that `.env` should be in the root
     directory of this repository.
 
-### Running the Application
+4. Pull the containers:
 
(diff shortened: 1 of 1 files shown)


sgsunder:
Merged manually in 79d0efc25b42d5c750743bd9c0d6c0d18df8e147

Thanks for the PR!

PR #504, opened 2022-09-10
support ftypiso6 file signature

w1kl4s:


neobooru:
For future reference: it looks like this is the MPEG-DASH format. Can't really find a source and the official spec costs money.

PR #534, opened 2022-12-30
Default upload tags and anonymous uploads

Fabricio20:
Hello,

I've taken the cherry-pick from [noirscape](https://github.com/noirscape)'s #468 , which was made by https://github.com/Hunternif/szurubooru/commit/035ce19788626c007619c004ef3b702f70ac262e and merged + tested it on my local environment, fixing an issue with tag name escaping as well.

Additionally, I added an option to mass upload posts anonymously, since if you are uploading a big mass of images it's annoying to mark all posts individually as anonymous. Due to there now being 4 checkboxes on the page, I've taken a bit of time to redesign the layout so it fits better.

image

These changes affect only the frontend and have all been tested on a local instance.

### DIFF
--- a/client/css/post-upload.styl
+++ b/client/css/post-upload.styl
@@ -15,6 +15,7 @@ $cancel-button-color = tomato
         &.inactive .skip-duplicates
         &.inactive .always-upload-similar
         &.inactive .pause-remain-on-error
+        &.inactive #common-tags,
         &.uploading input[type=submit],
         &.uploading .skip-duplicates,
         &.uploading .always-upload-similar
@@ -29,6 +30,9 @@ $cancel-button-color = tomato
             padding: 2em
             small
                 font-size: 60%
+
+    label[for=common-tags]
+        display: none !important
 
     input[type=submit]
         margin-top: 1em
--- a/client/css/post-upload.styl
+++ b/client/css/post-upload.styl
@@ -15,11 +15,13 @@ $cancel-button-color = tomato
         &.inactive .skip-duplicates
         &.inactive .always-upload-similar
         &.inactive .pause-remain-on-error
+        &.inactive .upload-all-anonymous
         &.inactive #common-tags,
         &.uploading input[type=submit],
         &.uploading .skip-duplicates,
         &.uploading .always-upload-similar
         &.uploading .pause-remain-on-error
+        &.uploading .upload-all-anonymous
         &:not(.uploading) .cancel
             display: none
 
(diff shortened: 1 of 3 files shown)


neobooru:
Hey, thanks for the work. I tried out some more changes for the UI. In replaced the input box with a TagInputControl (which gives tag colors, a list with a delete button, etc) and I added a collapsible expander. In the first version I only put the tags control in the expander, and in the second version I also put the options/checkboxes in an exapander.

Obviously we also could just not use any expanders. Please tell me what you think (and others please do so as well if you have an opinion about this).

<details>
  <summary>First version: one expander</summary>

firefox_2022_12_30_21_44_1059
firefox_2022_12_30_21_44_1060
</details>

<details>
  <summary>Second version: two expanders</summary>

firefox_2022_12_30_21_50_1062
firefox_2022_12_30_21_50_1061
</details>

Fabricio20:
Hey neo, thanks for the quick reply on this! I quite like your changes and (personally) I feel like the two dropdowns version looks better as it properly separates the options from the tags, as well as giving a proper section header for the user "Options" and "Common Tags", as opposed to my empty textbox without a label.

noirscape:
Personally I would recommend the second one. Feels the neatest and cleanest if we're gonna do the upload choices not on one/two rows anymore.

Ideal version would probably just be to add a new line and do two more in-line options but this looks fine too.

neobooru:
<details>
  <summary>Third version: flex wrap</summary>

firefox_2023_01_05_21_20_1068
firefox_2023_01_05_21_21_1069
</details>

<details>
  <summary>Fourth version: two cols</summary>

firefox_2023_01_05_21_26_1070
</details>

<details>
  <summary>Fifth version: three cols</summary>

firefox_2023_01_05_21_28_1071
</details>

Third version is the most *flex*ible when adding new items or on a different font-size. The other two quickly start looking ugly if you need two lines for the option name. This currently is not the problem (on default font-size), but maybe in the future or on other font-sizes?

I think the fifth version looks the best, but the rest is also okay. Opinions?

noirscape:
My personal preference is the fifth version out of all of these. Flex wrap reeks a little bit too much of "this is chosen because its pretty easy". 

That said, I can also imagine that if we ever need to add more options, we can always revise this input box as necessary depending on what looks the best. Generally speaking these options are something you consciously enable anyway.

Did you also check these on the small screen/mobile version? I doubt it's an issue, but if the font-size is different, then it might be useful to check if it still looks good there.

neobooru:
It currently resizes like this. It's not perfect, but the only alternative would probably be to just put them underneath each other as soon as it doesn't fit inside a 2x2 grid anymore.

https://user-images.githubusercontent.com/50623835/211850913-9bf8cfdd-bcdc-4c1a-b80f-d393e5e88d70.mp4

Source is available on [neobooru:feature-534](https://github.com/neobooru/szurubooru/tree/feature-534). I don't think I can add commits to this PR, so I'll probably manually squash and merge it.

PR #537, opened 2023-01-04
Poolpost nearby implementation

noirscape:
This is a draft to redo a component of #403 (which is seemingly abandoned for 2 years now and had some choice comments). Specifically, this PR aims to introduce the pool navigation elements to the post page.

Right now it is still in draft (I need to test this, write unit tests and add the frontend elements) but I'm making this draft PR to inform people on the status of the implementation.

I might choose to implement other elements of #403 in other PRs (although I'd probably do element 3 differently and not do element 2 at all, but we'll see on that one), but I really want this component functional.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -975,6 +975,43 @@ data.
 
     Retrieves information about posts that are before or after an existing post.
 
+## Getting pools around post
+- **Request**
+
+    `GET /post/<id>/pools-nearby`
+
+- **Output**
+
+    ```json5
+    [
+        {
+            "pool": <pool>,
+            "firstPost": <first-post>,
+            "lastPost": <last-post>,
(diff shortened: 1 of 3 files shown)


noirscape:
Okay sorry for all the extra commits; the backend stuff is now pretty much tested on my end. I'll write some formal tests later.

Next up is the frontend. I'm not *terribly* familiar with the frontend so this may take a bit.

noirscape:
And that should be most of the frontend code (wound up lifting the majority of it wholesale from #403). I'll write some unit tests later, and then it should be ready for review.

po5:
Undefined `$pool-navigator-header-background-color-darktheme` in `pool-navigator-control.styl`.

PR #552, opened 2023-02-21
Default upload tags and anonymous uploads (improved)

neobooru:


po5:
Would incorrectly check the anonymous box when duplicates were found. https://github.com/po5/szurubooru/commit/91b11c5b584444bf7637683773999997de01b442
The same querySelector led to the anonymous checkbox on individual images not actually doing anything. https://github.com/po5/szurubooru/commit/02c230ec9525a9ab2aa153e130f25691a11a642e

Giraffaman:
hi - are there still any plans to merge this? Having default tags when uploading would be a great addition.

KenwoodFox:
Oh! This would be super handy, auto-set tags and safety level?

PR #577, opened 2023-05-12
Add the ability to ban posts

Rebel4788:
I don't know if this is a feature this project wants but I intend to use it and you may like the feature. Feel free to dismiss.

This adds the ability to ban uploaded post files and prevent them from being uploaded again.

### DIFF
--- a/client/html/post_edit_sidebar.tpl
+++ b/client/html/post_edit_sidebar.tpl
@@ -108,7 +108,7 @@
             </section>
         <% } %>
 
-        <% if (ctx.canFeaturePosts || ctx.canDeletePosts || ctx.canMergePosts) { %>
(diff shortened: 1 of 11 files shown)


noirscape:
I like this in concept, seems very useful to comply with the DMCA and similar copyright laws for public boorus. That said, it seems to be missing a few endpoints:

* Listing banned sha1sums does not seem to be possible at the moment.
* There's similarly, no unban mechanism to lift a ban later down the line.

Would suggest adding an endpoint + frontend stuff for those two things.

Rebel4788:
Good points. I've added basic endpoints + frontend pages for listing and removing banned posts. Hopefully this is satisfactory.

I'm no expert in this technology stack so this may need a good look to make sure everything is implemented correctly.

PR #583, opened 2023-06-08
Added user-wide tag blocklist

### LINKED ISSUE #317: Feature request: User-based tag blacklisting
As a regular or higher-permission user, I would like to be able to give a list of tags that are considered "blacklisted" for that me. I would put those tags somewhere in my user settings, and when browsing posts, posts that are tagged with at least one of my blacklisted tags will not appear for me.

I would prefer if the blacklist were associated with the user account rather than localstorage or other browser-based persistence; if I log in to the szurubooru server from any computer anywhere, I'd like to have the same blacklist.

Is this something that would be considered within the scope of this project?

Lugrim:
Added a blocklist field to user settings. Any image containing any of this tag will be hidden to the user, unless they explicitly add it to their request.

Model wise, it isn't perfect (Text field), but hopefully good-enough.

Intends to fix #317 

Comments and reviews are greatly appreciated !

### DIFF
--- a/client/css/comment-control.styl
+++ b/client/css/comment-control.styl
@@ -127,6 +127,10 @@ $comment-border-color = #DDD
             color: mix($main-color, $inactive-link-color-darktheme)
 
 .comment-content
+    p
+        word-wrap: normal
+        word-break: break-all
+
     ul, ol
         list-style-position: inside
         margin: 1em 0


bakedhorse:
This looks and works great as far as I have tested. However I do have 2 suggestions. Just personal nitpicks, but the second one should be easier than the first.

1. I'm not a huge fan of the blocklist being a text field, though the pop up list when typing is nice to have. Maybe if it was in its own tab on the account page and then displayed the tags the same way as they do on the post page, along with a text field to add tags to said tags.
Screenshot 2023-07-14 082159

2. A configurable default list would be nice, another option in the config.yml (option name in image is just example)
Would allow the owner/admins to create a list of already blocklisted tags for new users.
Screenshot 2023-07-14 082501

Soblow:
We're working on the first suggestion, but if you want, the second one (default server blacklist) is ready for testing!

bakedhorse:
The second one is working great!  Could it be applied to anonymous users as well? 
Maybe another toggle named like 'enforced_blacklist_anonymous', allows owners/admins to decide if they want the default blacklist enforced onto anonymous users.

Take your time for the first suggestion, its nothing that needs to be rushed. Better to put out better code than rushed 'good enough' code.

Soblow:
Sorry for the lack of activity, the patch is sitting on my computer since a month now and I totally forgot about it.

I completed a rewrite of my blocklist patch to use a separate DB table, and also implement both of the changes asked by @bakedhorse. I'm waiting for a review from @Lugrim before updating the PR here.

I have some weird issues with tests so the patches only include fixes for existing ones, and 2 "new" tests, but most of the other originally planned tests aren't written.

Soblow:
We had to reopen a new MR, but this is basically the continuation of work done here...

PR #593, opened 2023-07-22
Add ability to configure allowed file types

pbf:
This PR adds ability to configure allowed file types via `config.yaml` file. I went for using MIME types identifiers rather than extensions because it maps better to existing file type handling in szurubooru. The files that are of unallowed types are rejected with `szuruburu.func.posts.InvalidPostContentError`, same as with unsupported files but with different error message.

### DIFF
--- a/client/js/api.js
+++ b/client/js/api.js
@@ -100,6 +100,10 @@ class Api extends events.EventTarget {
         return remoteConfig.contactEmail;
     }
 
+    getAllowedExtensions() {
+    	return remoteConfig.allowedExtensions;
+    }
+
     canSendMails() {
         return !!remoteConfig.canSendMails;
     }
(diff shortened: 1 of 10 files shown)


sticknuder:
When I tried to merge this into my custom Szurubooru fork, I couldn't figure out how to allow my own custom filetype. 

I'm wondering if this PR could help me allow arbitrary binary files with a specific set of possible extensions. (.nodes, .stknds, and .nodemc)

I understand that they wouldn't be preview-able, but I would like to allow them nonetheless.

pbf:
@sticknuder Sorry for late answer, but I hadn't visited this PR in a while.

No, I'm afraid this PR won't help you with implementing this kind of functionality. It adds option to restrict file types that are allowed on a given szurubooru instance, but it doesn't extend the file format support in szurubooru itself. Adding such support may be tricky, since szurubooru seem to throw an error in `func.posts.update_post_content` if new post isn't an image, video or flash animation. I suppose one would need to code the handling of such generic types as you describe.

sticknuder:
@pbf thanks for the response, but don't worry in the end I was able to do it 😁👍

Issue #486, opened 2022-05-27
-source:* does not work, or: some way to search for posts without sources


404Fox:
Negated source does work but only on images that already have sources. 

`-source:twitter` does work.
`source:*` does work. 
`-source:*` result returns "No data to show"

this is probably so that `-source:twitter` doesn't show all unsourced posts alongside that result.

alnmy:
Having same issue, could this be looked into?
Using `-source:*...*` seems to exclude all posts which don't have a source, which isn't the behaviour which I think most people would want as most people would instead want it to show all posts which both don't have a source and don't match the source being negated.

Issue #619, opened 2023-12-03
Negated source filter excludes posts with no source


halbrd:
A source filter such as `source:*reddit.com*` works as expected, returning posts with "reddit.com" in the source field.

However, try negating the source filter: `-source:*reddit.com*`.

Expected function: return the inverse of the above - posts that don't have "reddit.com" in the source field (including posts with no source content at all).

Actual function: it only returns posts with something in the source field (and don't include "reddit.com") - posts with no source aren't returned here.

This seems unintuitive, and more importantly, there doesn't seem to be a way to achieve the expected function.

zaidalkaz:
This issue happens with "upload" as well - posts uploaded by anonymous users do not show up when searching for a negated filter e.g. "-upload:zaidalkaz". I suspect any "empty" field fails to show when searching in this way

PR #627, opened 2023-12-16
server: make negated string filters include null values

halbrd:
This appears to produce the expected behaviour described in issue #619.

I would like a sanity check from someone familiar with the codebase though.


hujle:
I have applied this pull request to my instance, but it's still giving me no results when I'm searching `-source: `.

halbrd:
Really? I just checked `source:` and `-source:` on my dev instance and they return 'any source' and 'not any source' as you would expect.

po5:
Closes #486 too.

Issue #449, opened 2021-11-12
"Could not fetch basic configuration from server" After restart.


AveryLucas:
I did not change any settings in config and everything works great on first run but after a restart of the computer, restarting Szurubooru gives this error.

Not sure how to prevent this from happening.

sgsunder:
Can you post log files?

If you are running on Windows 10, are you using the Docker install? The Docker containers we use are Linux-specific, so you need to make sure that your Docker installation is treating it like a Linux container and not a Windows one.

Mrspirits:
Hey, I have the same issue as well.

Here's the error I keep getting in the logs from docker:

`szuru-client-1  | 172.18.0.1 -> GET /api/info HTTP/1.1 [504] - referer: http://127.0.0.1:8080/ -
szuru-client-1  | 2021/12/04 20:07:32 [error] 11#11: *1 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 172.18.0.1, server: , request: "GET /api/user/admin?bump-login=true HTTP/1.1", upstream: "http://172.18.0.3:6666/user/admin?bump-login=true", host: "127.0.0.1:8080", referrer: "http://127.0.0.1:8080/"`

I use Docker on Windows 11 and I'm assuming the issue is that there are a bunch of things up on the booru and that's slowing it down and making it time out.

neobooru:
> It does fix itself after a while but it takes like 10 minutes to finally start working.

Does the server container show you any output? The server does run some tasks at startup, but they should show up in the logs.

Mrspirits:
It lists a bunch of tags out then comes up with this:

`szuru-server-1  | INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
szuru-server-1  | INFO  [alembic.runtime.migration] Will assume transactional DDL.
szuru-server-1  | Starting szurubooru API on port 6666`

and after that it keeps timing out with the error from before until it eventually does this:

`szuru-client-1  | 2021/12/06 12:50:08 [error] 10#10: *25 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 172.18.0.1, server: , request: "DELETE /api/user-token/admin/03d7c379-1220-41e3-b6a4-f1a8db1b6eba HTTP/1.1", upstream: "http://172.18.0.3:6666/user-token/admin/03d7c379-1220-41e3-b6a4-f1a8db1b6eba", host: "127.0.0.1:8080", referrer: "http://127.0.0.1:8080/home"
[...]

PR #638, opened 2024-03-04
Adding option to disable the _get_disk_usage

Theenoro:
It's solving this issue #449 in https://github.com/rr-/szurubooru/issues/449#issuecomment-1416593833

I've added the configuration part and the python part to disable the function (_get_disk_usage) in the info api.
It's useful to boost up the page speed in the trade of the disk usage indicator.

### DIFF
--- a/server/config.yaml.dist
+++ b/server/config.yaml.dist
@@ -177,3 +177,9 @@ privileges:
 ## usage: schema://user:password@host:port/database_name
 ## example: postgres://szuru:dog@localhost:5432/szuru_test
 #database:
+
+
+## get the data dir usage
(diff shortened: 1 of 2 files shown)


phil-flip:
Thank you for the PR.
I don't code Python, but looking at other code lines it might be more standardized, when using true/false as the value and removing `== 1`.

Theenoro:
> Thank you for the PR. I don't code Python, but looking at other code lines it might be more standardized, when using true/false as the value and removing `== 1`.

yes, but I also found something in the options with yes/no. I actually dunno why I used 1/0 instead of true/false but I've changed it.

PR #677, opened 2024-09-01
If attempting to reverse-search using a PSD, convert to PNG first

Deer-Spangle:
Currently, if attempting to reverse-search and uploading a PSD, the backend will take a long time (sometimes over a minute) trying to open the PSD before failing and using only exact file matches. 

This is fine, as long as the time is less than 1 minute, but if it's longer than that, the API times out.

Turns out converting to a PNG is much faster than trying to open a PSD, so this fixes that and gives some rough image check


Deer-Spangle:
Sorry, meant to make this to PR my own fork!

PR #687, opened 2024-10-24
Support reading env vars on Podman

recursiveribbons:
When a container is started on Podman, it creates a file at `/run/.containerenv` instead of `/.dockerenv`. Everything else otherwise runs identically. Detecting this file thereby gives it support for Podman and properly load the environment variables.


recursiveribbons:
Superseded by #689

PR #689, opened 2024-11-03
server/config: generalize container support

zakame:
Allow running in Kubernetes, podman, and LXC, besides plain docker-compose,
without having to fake out /.dockerenv in non-Docker environments.


zakame:
Thanks @neobooru ! 🎉 🚀

PR #696, opened 2024-11-16
Fixed the tag suggestions being shown as [object Object] when added m…

Anbosuki:
Hello,
when manually adding a tag suggestion to a tag under the tags edit page you cannot actually see it when you want to use it. If you click on the tag while editing a post, you simply see `[object Object]`, which isn't helpful of course.

It can be simply fixed by accessing `suggestion._origName`. I tested it and it works without any problems. Even when adding multiple suggestions to an tag it doesn't cause any issues.


po5 (approved):

po5:
There's another instance of this, I submitted a PR to your branch: https://github.com/Anbosuki/szurubooru/pull/1

Issue #430, opened 2021-08-11
Privileges for safety


JacksonSparkle:
Can there be privileges to set the posts:list for the different safety flags?

PR #714, opened 2025-02-16
Privelege to list & view unsafe posts

Hunternif:
_Use case: I want to show off the awesome szurubooru to my friend, but I don't want them to see my nsfw posts XD_

This PR adds 2 privileges:
```
'posts:list:unsafe':    regular
'posts:view:unsafe':    regular
```

The `view` privilege restricts viewing posts via `post_api`.
The `list` privilege _filters out_ any "unsafe" posts from your query, for all queries.
Adds unit tests too.

### DIFF
--- a/server/config.yaml.dist
+++ b/server/config.yaml.dist
@@ -97,6 +97,7 @@ privileges:
     'posts:create:anonymous':       regular
     'posts:create:identified':      regular
     'posts:list':                   anonymous
+    'posts:list:unsafe':            regular
     'posts:reverse_search':         regular
     'posts:view':                   anonymous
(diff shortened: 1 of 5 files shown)


hujle:
That's a good feature, but I wish it wasn't pinned to unsafe category and could be applied to any of them.

On that note, an extra "hidden" category with these permissions applied to it would be better solution.

po5:
The post listing change leads to cache collisions, I submitted a PR with a fix explaining the issue. https://github.com/Hunternif/szurubooru/pull/2

po5:
Closes #430.

Issue #395, opened 2021-05-03
Is there a better way to iterate on development?


Jazzepi:
I had some basic questions about developing on the booru codebase.

It seems like the only way to iterate on the booru is to rebuild the docker image(s) every time you have a change, even if it's something simple like an HTML template or a CSS style.

https://github.com/rr-/szurubooru/wiki/Customizing-UI-Colors

Is there any way to rebuild the client facing stuff without building the entire docker image and starting the container anew? I saw a watch script in the client directory but that doesn't seem to help if you've got everything baked into docker images.
Same questions for the python code in the web server. I assume that since it's interpreted all you would need is a way to edit the mounted files. Is something like that available?
Is there information on how to setup debugging for the python running inside the docker?
Are there any tests setup for the system? If so, how do we run them?

Ruin0x11:
I agree that the onboarding/development process is way too hard for new contributors. I had to hack together some scripts and an Ngnix config to develop outside of Docker.

I'm able to run a development instance on my box (without Docker) with these steps. My distro is Arch Linux.

1. Install postgresql and nginx.

```bash
yay -S postgresql nginx
```

2. Create the database and add the szurubooru user and database.

```bash
# Only needed if installing postgres for the first time
sudo -iu postgres initdb -D /var/lib/postgres/data

sudo systemctl start postgresql
sudo -iu postgres createuser szuru -s
sudo -iu postgres psql -c "ALTER USER szuru PASSWORD 'dog';" # set a password to be used later
sudo -iu postgres createdb szuru
```

3. Edit `server/config.yaml` and scroll to the bottom. Update these fields:

```yaml
## ONLY SET THESE IF DEPLOYING OUTSIDE OF DOCKER
debug: 1 # generate server logs?
show_sql: 1 # show sql in server logs?
data_url: data/
data_dir: /home/hiro/build/work/szurubooru/client/public/data
## usage: schema://user:password@host:port/database_name
## example: postgres://szuru:dog@localhost:5432/szuru_test
database: postgres://szuru:dog@localhost:5432/szuru
```

`/home/hiro/build/work/szurubooru/` is my checkout of the Git repo. `szuru:dog` is the username/password. `localhost:5432/szuru` is the host/database name.

4. Install the Node dependencies and build the `vendor.min.js` bundle for the first time. Subsequent bundles will skip this step to save time.

```bash
cd ./client
npm install
npm run build
```

5. Install the Python dependencies. Virtualenv should be installed on the system.

```bash
[...]

PR #719, opened 2025-03-05
build: create docker-compose.dev.yml so dev build iteration doesn't take forever

swomf:
Earlier, [neobooru started work](https://github.com/neobooru/szurubooru/blob/docker-development-setup) to create a `docker-compose.dev.yml` for rapid hot-reloadable development in response to #395.

This commit should cleanly merge `docker-development-setup` into the current `master`, adding and fixing anything missing while keeping the scope of the change as small as possible.

Compared to downstream's branch, this commit
* Excludes extraneous changes such as
    * Any formatting
    * The use of [deprecated](https://docs.docker.com/reference/compose-file/version-and-name/)/ineffectual top-level `version:` in composer files
* Supports controlling `$THREADS`
* Respects currently used `FROM` statements, yadayada

## limitations

The volume mounting setup during the `client, target development` Docker build step causes `node_modules` to disappear. To ensure that `node_modules` are present at runtime I simply run `npm i` before any given `npm run watch` in `client/docker-start-dev.sh`.

This commit doesn't add docs yet :-) but it should be along the lines of a simple

```bash
docker compose -f ./docker-compose.dev.yml up
```

### DIFF
--- a/client/.dockerignore
+++ b/client/.dockerignore
@@ -1,4 +1,5 @@
 node_modules/*
+public/
 Dockerfile
 .dockerignore
 **/.gitignore
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,3 +1,26 @@
+FROM --platform=$BUILDPLATFORM node:lts-alpine as development
+WORKDIR /opt/app
+
+RUN apk --no-cache add \
+    dumb-init \
+    nginx \
+    git
+
+RUN ln -sf /opt/app/nginx.conf.docker /etc/nginx/nginx.conf
+RUN rm -rf /var/www
+RUN ln -sf /opt/app/public/ /var/www
+
+COPY package.json package-lock.json ./
+RUN npm install
+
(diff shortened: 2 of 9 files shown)


neobooru:
Looks mostly good, but in client/build.js and client/js/main.js you need to update the WebSocket.Server port from 8080 to 8081.

swomf:
The two lines should be fixed now. I squashed the addition for a cleaner merge though.

```diff
# client/build.js 
function watch() {
-    let wss = new WebSocket.Server({ port: 8080 });
+    let wss = new WebSocket.Server({ port: environment === "development" ? 8081 : 8080 });
```

```diff
# client/js/main.js
if (config.environment == "development") {
-    var ws = new WebSocket("ws://" + location.hostname + ":8080");
+    var ws = new WebSocket("ws://" + location.hostname + ":8081");
```

swomf:
Hey, is this revision sufficient?

Issue #542, opened 2023-01-21
Markdown parsing of post numbers like @42 breaks if using a different base directory


recordcrash:
Using @43 resolves to https://homestuck.net/post/43, despite the base URL of my self hosted docker wiki being https://homestuck.net/resources/booru/ (`BASE_URL=/resources/booru/`). 

It should resolve to https://homestuck.net/resources/booru/post/43. I don't think this should be particularly hard to fix, but it'd be pretty useful.

PR #738, opened 2025-04-08
client/markdown: small fixes and improvements

G1org1owo:
Simple changes to markdown engine to allow use of an `[icon]...[/icon]` tag to insert an icon alongside a link.

This was inspired by danbooru artist pages, which have an easily recognizable icon on the left of the artist's social links, making it way easier to find the links one's looking for. Manually doing this for each artist would be a very tedious matter on a szurubooru instance, so I introduced this tag.

For example, putting `[icon]https://www.youtube.com/watch?v=dQw4w9WgXcQ[/icon]` in a markdown section gets it rendered as
[https://www.youtube.com/watch?v=dQw4w9WgXcQ](https://www.youtube.com/watch?v=dQw4w9WgXcQ) https://www.youtube.com/watch?v=dQw4w9WgXcQ.

### DIFF
--- a/client/html/help_comments.tpl
+++ b/client/html/help_comments.tpl
@@ -26,6 +26,10 @@
             <td><code>[sjis](´･ω･`)[/sjis]</td>
             <td>adds SJIS art</td>
         </tr>
+        <tr>
+            <td><code>[icon]https://youtube.com[/icon]</td>
+            <td>adds the site icon next to the link</td>
+        </tr>
     </tbody>
 </table>
 
--- a/client/html/help_comments.tpl
+++ b/client/html/help_comments.tpl
@@ -11,6 +11,10 @@
             <td>links to tag &ldquo;Dragon_Ball&rdquo;</td>
         </tr>
         <tr>
+            <td><code>?Hatsune_Miku</code></td>
+            <td>links to tag page for &ldquo;Hatsune_Miku&rdquo;</td>
+        </tr>
+        <tr>
             <td><code>+Pirate</code></td>
             <td>links to user &ldquo;Pirate&rdquo;</td>
         </tr>
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -110,6 +110,15 @@ class StrikeThroughWrapper extends BaseMarkdownWrapper {
     }
 }
 
+class FaviconWrapper extends BaseMarkdownWrapper {
+    preprocess(text) {
+        return text.replace(
+            /\[icon\]((?:[^\[]|\[(?!\/?icon\]))+)\[\/icon\]/gi,
+            "[![$1](https://www.google.com/s2/favicons?domain=$1)]($1) $1"
+        );
+    }
+}
+
 function createRenderer() {
     function sanitize(str) {
         return str.replace(/&<"/g, (m) => {
@@ -155,6 +164,7 @@ function formatMarkdown(text) {
         new SpoilersWrapper(),
         new SmallWrapper(),
         new StrikeThroughWrapper(),
+        new FaviconWrapper(),
     ];
     for (let wrapper of wrappers) {
         text = wrapper.preprocess(text);
@@ -181,6 +191,7 @@ function formatInlineMarkdown(text) {
         new SpoilersWrapper(),
         new SmallWrapper(),
         new StrikeThroughWrapper(),
+        new FaviconWrapper(),
     ];
     for (let wrapper of wrappers) {
         text = wrapper.preprocess(text);
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -51,17 +51,6 @@ class TildeWrapper extends BaseMarkdownWrapper {
     }
 }
 
-// prevent ^#... from being treated as headers, due to tag permalinks
-class TagPermalinkFixWrapper extends BaseMarkdownWrapper {
-    preprocess(text) {
-        return text.replace(/^#/g, "%%%#");
-    }
-
-    postprocess(text) {
-        return text.replace(/%%%#/g, "#");
-    }
-}
-
 // post, user and tags permalinks
 class EntityPermalinkWrapper extends BaseMarkdownWrapper {
     preprocess(text) {
@@ -158,7 +147,6 @@ function formatMarkdown(text) {
     let wrappers = [
         new SjisWrapper(),
         new TildeWrapper(),
-        new TagPermalinkFixWrapper(),
         new EntityPermalinkWrapper(),
         new SearchPermalinkWrapper(),
         new SpoilersWrapper(),
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -116,7 +116,7 @@ class FaviconWrapper extends BaseMarkdownWrapper {
     preprocess(text) {
         return text.replace(
             /\[icon\]((?:[^\[]|\[(?!\/?icon\]))+)\[\/icon\]/gi,
-            "[![$1](https://www.google.com/s2/favicons?domain=$1)]($1) $1"
+            '<a href="$1"><img src="https://www.google.com/s2/favicons?domain=$1"> $1</a>'
         );
     }
 }
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -55,12 +55,16 @@ class TildeWrapper extends BaseMarkdownWrapper {
 class EntityPermalinkWrapper extends BaseMarkdownWrapper {
     preprocess(text) {
         text = text.replace(
-            /(^|^\(|(?:[^\]])\(|[\s<>\[\]\)])([+#@][a-zA-Z0-9_-]+)/g,
-            "$1[$2]($2)"
+            /(?<=(?<!\])\(|[\[<])([+#@?][^\s%#+/]+)(?=[\)\]>])/g, "[$1]($1)"
         );
+
+        text = text.replace(
+            /(?<![\(\[<]|%%%)([+#@?][^\s%#+/]+)(?![\)\]>])/g, "[$1]($1)"
+        );
+
         text = text.replace(/\]\(@(\d+)\)/g, "](/post/$1)");
         text = text.replace(/\]\(\+([a-zA-Z0-9_-]+)\)/g, "](/user/$1)");
-        text = text.replace(/\]\(#([a-zA-Z0-9_-]+)\)/g, "](/posts/query=$1)");
+        text = text.replace(/\]\(#([^\s%+#/]+)\)/g, "](/posts/query=$1)");
         return text;
     }
 }
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -57,13 +57,13 @@ class EntityPermalinkWrapper extends BaseMarkdownWrapper {
         super();
         this.getPrettyName = getPrettyName || ((text) => text);
     }
+    unescape(text) {
+        return text.replace(/\\([^#@+/])/g, "$1");
+    }
     preprocess(text) {
         text = text.replace(
-            /(?<=(?<!\])\(|[\[<])([+#@?][^\s%#+/]+)(?=[\)\]>])/g, "[$1]($1)"
-        );
-
-        text = text.replace(
-            /(?<![\(\[<]|%%%)([+#@?][^\s%#+/]+)(?![\)\]>])/g, "[$1]($1)"
+            /(?<![a-zA-Z0-9])([#+@?](?:[a-zA-Z0-9_-]|\\[^#@+/])+)/g,
+            (_, entity) => `[${this.unescape(entity)}](${this.unescape(entity)})`
         );
 
         text = text.replace(/\]\(@(\d+)\)/g, "](/post/$1)");
--- a/client/js/util/markdown.js
+++ b/client/js/util/markdown.js
@@ -65,6 +65,9 @@ class EntityPermalinkWrapper extends BaseMarkdownWrapper {
         text = text.replace(/\]\(@(\d+)\)/g, "](/post/$1)");
         text = text.replace(/\]\(\+([a-zA-Z0-9_-]+)\)/g, "](/user/$1)");
         text = text.replace(/\]\(#([^\s%+#/]+)\)/g, "](/posts/query=$1)");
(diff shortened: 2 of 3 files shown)


po5:
There's a bug in cebb9fbdff6aeb95c67fd7f74872aa5badf022e4 that merges links with adjacent text.
Example:
image
Rendered:
image
I believe this is why `SearchPermalinkWrapper` (which doesn't exhibit this issue) uses HTML instead of piggybacking off the Markdown parser.

For 20ea6bbb491bc2e37a209e10a46786e7fbce6213, can you explain the advantage of your double lookaround regex, if any? Does it match in fewer steps? Does it fix an edge case? Why two replaces? Surely the entire regex didn't need rewriting to accommodate the looser tag character set.

9c8ceb5c0255d024263164dbe27c28ca267d2925: We don't turn underscores into spaces for any other formatting, so I don't think it's appropriate for the wiki links to do that unconditionally. But I'm not opposed to updating `#@+` to use `getPrettyName()` to respect user settings.
This commit also mangles any links that contain a query string.
Example:
image
Rendered:
image

I don't see any issue with 9a7c1f232271b832c69154f05eea5303be9995ad, I think your reasoning is correct. It's better than my solution (5d1dbb291c60df05f990b7f818ddd1fd609d1ec0) since the code served no purpose.

po5:
It might be worth looking at how other boorus do it. Danbooru has pretty unambiguous `{{tagname1 tagname2}}` syntax for search. https://danbooru.donmai.us/wiki_pages/help%3Adtext

G1org1owo:
> I can come up with more rules like a tag #negima!. How do we know if the user was referring to a tag with an exclamation mark, or if that was part of their text? In a world with free-form tag names I don't think it makes sense to write a magic parser with lots and lots of rules to make it more usable.
I think requiring escaping for all "special" characters is the best solution, even if backslashes are annoying to type. It could be eased by adding tag autocomplete when a user is typing after a #.
The only characters we'd allow unescaped would be [a-zA-Z0-9_-]. Maybe allow CJK characters too.

I feel pretty stupid for not thinking of just escaping characters before, I was too focused on doing it with a regex alone. This might actually be doable. The regex `/([#+@?](?:[a-zA-Z0-9_-]|\\[^#@+/])+)/gu` works with all alphanumeric characters and allows escaping for all valid non-alphanumeric characters.

I tried allowing CJK characters without escapes with the regex `/([#+@?](?:[a-zA-Z0-9_-\p{Han}]|\\[^#@+/])+)/gu` but I received a warning it might not work on all browsers on [regexr.com](https://regexr.com) and it didn't seem to actually match anything.

> I don't want to scope creep your PR, but as explained I don't think solving this issue even for the default allowed character set is possible without rethinking how tags work everywhere. It's already very inconsistent. Might be best left for another day.

I am not aware of other places with problematic behaviour, please do let me know if this proposed solution has other downsides I don't see right now.

> escapeTagName needs to be used here and in a lot of other places, that will be an improvement.

I'll look into it as I plan to use it with this escape-style regex, though I'm not sure about the other places you mentioned.

> It might be worth looking at how other boorus do it. Danbooru has pretty unambiguous {{tagname1 tagname2}} syntax for search. https://danbooru.donmai.us/wiki_pages/help%3Adtext

I was quite in favour of using unambiguous tags such as `[tag]tag_name[/tag]` and Danbooru syntax has the advantage of allowing referencing of multiple tags for a search (which I'd rather leave for a later issue right now), but since changing this would break previous uses (unless we create some kind of DB migration that updates existing comments/descriptions but I don't really like that) I'd much rather avoid it unless it's the only reasonable way to do this.

I'll push a couple commits trying this later today.

po5:
> I received a warning it might not work on all browsers on [regexr.com](https://regexr.com/) and it didn't seem to actually match anything.

It's been supported in all browsers [for close to 10 years now](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#browser_compatibility), that's not a concern. And everything gets transpiled, but more on that below...

Your new regex still mangles urls with a trailing slash/questionmark + hash, like `https://example.com/#hello`. It also doesn't work when a comment begins with a tag shortlink.
I've submitted a PR https://github.com/G1org1owo/szurubooru/pull/6 to your fork that allows accented and CJK characters without escaping, allows escaped #@? in tag names, and uses the much more explicit rule of requiring a separator character before the #@?.

[regular-expressions.info](https://www.regular-expressions.info/shorthand.html) claims:
> In flavors that support Unicode, `\s` normally includes all characters from the Unicode "separator" category. Java and PCRE are exceptions once again. But JavaScript does match all Unicode whitespace with `\s`.

So I went with `\s` instead of the more esoteric `\p{Z}` that can't even be parsed by the regjsparser library that babelify uses.
Like in the original code I had to capture the separator because a positive lookbehind `(?<=\s|^)`, because of the same bad parser issues.
And-- after much wrestling with the awful transpilation capabilities of our current setup, I gave up and pre-processed it with regexpu. Sorry about the ugly mess that's now in the source. That can be fixed the day that we update all our libraries.
I had to include the pre-transpiled version of `\s` too because we reuse some patterns as strings and the transpiler isn't smart enough to identify regex patterns in strings.

In making the final replace for `/posts/query=` stricter, you broke the original intent of the code to allow links like `[display text](#mytag)` to link to tag search. And even mangles it into `<a href="%5B#mytag%5D(/posts/query=mytag)">display text</a>`.

> Danbooru syntax has the advantage of allowing referencing of multiple tags for a search (which I'd rather leave for a later issue right now)

We already have this as [search][/search]. Speaking of, the existing `SearchPermalinkWrapper` is also easy to break, e.g. with `[search]test #oops[/search]` producing this:
image
It's not quite fixed in my PR, the search link still gets messed up because the parsers aren't aware of each other, but it at least doesn't look as broken.
image

Post ID linking was also jank, you were allowed to have anything after @ and it would make into a link that would get replaced in the first step but not the 2nd, leading to a dead end link when the user didn't even mean to make one.
Also all permalinks were wrong if you weren't hosting under a subdirectory (cf. #542) due to having a hardcoded slash at the beginning.
Both fixed in my PR.

I want to work on the autocomplete system to make it provide faster feedback, and be embeddable in textareas like comments with automated escaping on selection. I might not dive into that for a while though, so don't see it as a blocker for this PR.

po5:
I should note that the reason this works without having to unescape the "display text" part of the Markdown links, is because marked.js gets rid of them. But it won't get rid of escapes that were pointless, so `test\lol` will still render with a visible backslash, even if we ignore it in the actual search link.
I haven't implemented display text cleanup because I can't find a way to parse links that reproduces all the edge cases of marked.js's behavior for nested/malformed tags, so there's a chance we'd be messing stuff up.
It should be easier to match the behavior the other way around, making links match the displayed text. That way there are no surprises, what you see is what you get. Would just have to split up the escape regex into its own string so we can use just the valid character pattern on its own.

po5:
When we updated marked.js to v0.3.9 in 59d8b0d4c579771d3bfa2f757fe3498f3de3ddfe we gained support for `~~strikethrough~~` syntax, and when we updated past v0.5.0 in 6f549cf2db3b898c218488d19c7933e6d5cd9da9 we now had support for single tilde `~strikethrough~` as well.
Our `StrikeThroughWrapper` and `TildeWrapper` are now completely redundant and can be removed. This is nice because `TildeWrapper` was also getting in the way of permalinks with our new syntax like `#\~test\~`.
We are clear of `*` and `~` Markdown handling clashing with our new syntax, because we require escaping for these characters to be recognized as part of a tag. The same isn't true for `_`, which is allowed without escaping and renders Markdown when used in a tag like `#_test_` or `#__test__`. These styling rules also apply to all three problematic characters inside `[search]` tags. We need to properly escape them.

I knew fixing Markdown code would require a lot of consideration hahaha we're deep in it.

po5:
> Alternatively, it may be possible to set a custom handler for links in `createRenderer()` the same way we do for images. I don't know how robust that can be.

I implemented it. This allows us to strip unnecessary escapes, but all other issues remain.
We still get en/em dashes, and italics/bold from underscores. Strikethrough still possible in `[search]` tags.
This is because they get processed before reaching our custom function.
```js
    renderer.link = (href, title, text) => {
        let match;
        if ((match = /^@(\d+)$/.exec(href))) {
            href = uri.formatClientLink("post", match[1]);
        }
        else if ((match = /^([#+?])(.+)$/.exec(href))) {
            const prefix = match[1];
            const capture = match[2].replace(/\\(.)/g, "$1");
            text = text.replace(/\\(.)/g, "$1");
            if (prefix === "#") {
                href = uri.formatClientLink("posts", { query: uri.escapeTagName(capture) });
            } else if (prefix === "+") {
                href = uri.formatClientLink("user", capture);
            } else if (prefix === "?") {
                href = uri.formatClientLink("tag", capture);
            }
        }
        const titleAttr = title ? ` title="${escapeHtml(title)}"` : "";
        return `<a href="${escapeHtml(href)}"${titleAttr}>${text}</a>`;
    };
```
But it's possible to use marked's lexer and modify the values before feeding them into the final parser, this should work.
```js
    const tokens = marked.lexer(text);
    console.log(tokens);
    text = marked.parser(tokens, options);
```
It gives us links with their original text before any mangling:
image

po5:
I updated my PR to your fork https://github.com/G1org1owo/szurubooru/pull/6 with the lexer manipulation approach, we finally have something that can be trusted to display tags properly regardless of what weird characters or Markdown-like syntax they may contain :tada:
I had to mix the HTML escaping fix from #733 because it's required to do everything correctly.
Let me know how it works for you, and please try to break it any way you can.

It's an amazing achievement to have all of these working and linking to proper usable searches (:
```
[search]__test__ [notalink](yay) #not_a_tag *yay* [/search]

#__test__\ \[notalink\]\(yay\)\ \#not_a_tag\ \*yay\*\ 

#\"quoted\ search\"

#1\.\.
```
image

po5:
> The only solution I can see is to move all processing from `EntityPermalinkWrapper` into lexer stuff

Done. It was not trivial at all, but I think I successfully retained all current behavior.

Things that were not obvious I might get asked about if this PR ever gets reviewed:
- The move away from simple parsing is to avoid a lot of bugs that come with it.
- `parseLinksOnly()` exists because we need marked.js to linkify inline links like `www.google.com/#test` before processing entity shortlinks.
- The special handling of `br` is required to preserve line breaks.
- We strip the top level `p` tag. I originally went with `parseInline` since it doesn't generate extra stuff, but it collapses everything as if it was on a single line and causes issues with italics Markdown.
- `flattenTokens()` exists because marked.js creates single-character `"escape"` lines when a backslash is used. We need to work on whole strings when processing links.
- There is some similar logic in `modifyLink()` and `modifyTokens()`. One deals with parsed links that were already in the document (manually inserted links like `[check this out](#tag)`), the other with loose links like `#lonely_tag`. We can't piggyback off Markdown for loose links like we used to, because we're already past the lexing step. If we did it, it would require an extra render+lexer step. It's unnecessary, we already have everything we need.
- This means there is some shared logic. It's in `resolveEntity()`. I don't think the code can be deduplicated much more than this.
- `smartypants` functionality had to be reimplemented on our side because it's otherwise introduced by the marked.js lexer way too early for us to preserve tag names.
- The usage of e.g. `marked.Parser.parseInline` vs `marked.parseInline`, is because the former takes in tokens as opposed to text.

Known issues:
- When `[search]` tags or html tags e.g. `<test>` are inside a code block (` ``` `), we get doubly-encoded entities. `[search]` tags can be fixed by moving them to token logic. For special html characters it will require another careful look at the order of escapes. This is the only known area where the escaping behaves poorly.
- Links that contain spaces `[link](#with space)` are not handled by marked.js, and since we now rely on its tokenizer, trying to escape the space with a backslash doesn't help. Spaces (in the url part) hard-break any link. I don't think anyone will ever care, and you can just copy the full url to the page as a workaround since it will be url-encoded with `%20`.

G1org1owo:
> In making the final replace for /posts/query= stricter, you broke the original intent of the code to allow links like [display text](#mytag) to link to tag search. And even mangles it into <a href="%5B#mytag%5D(/posts/query=mytag)">display text</a>.

My bad, I misinterpreted the original code and thought the intent was to preserve URI fragment links for some reason.

> I updated my PR to your fork https://github.com/G1org1owo/szurubooru/pull/6 with the lexer manipulation approach, we finally have something that can be trusted to display tags properly regardless of what weird characters or Markdown-like syntax they may contain 🎉
I had to mix the HTML escaping fix from https://github.com/rr-/szurubooru/pull/733 because it's required to do everything correctly.
Let me know how it works for you, and please try to break it any way you can.

I read all of your commits so far and it mostly seemed to work for me, altough it's become quite more complex lol. Great work, really. I'll try breaking it with anything that comes to mind 👍

> I can spawn this off into a new PR to the main repo, if you can trim down your PR to just the [icon] tag.
I'll then clean up the commits and it will also allow me to work on tag autocomplete for comment inputs.

Sure, would you rather I trim on this one or should I close this one and open a new one just to keep it clean?

> Done. It was not trivial at all, but I think I successfully retained all current behavior.

Once again, you have done a great job and while it is definitely a complex module now I believe it's still quite understandable.

po5:
Links with spaces: Turns out it's already possible with this Markdown syntax `[test](<#with space>)` which has the benefit of being compatible with titles `[test](<#with space> "link title")`. This is nice syntax, So I added support for it in shortlinks e.g. simply `<#with space>`. It's much easier to type than `[search]with space[/search]`, and the only character that requires escaping is `>`.

Blockquotes used to be broken. Now fixed.
Shortlinks like `#__test__` would still trigger styling. That's fixed.
I unified processing of `[search]` tags with the rest, this does mean they no longer display with an extra `<code>` element, and are prefixed with `#`. But I don't think anyone will be bothered by the change.

I decided to not fix shortlinks inside ` ``` ` code blocks. (It's always been broken anyway)
I may make shortlinks work inside bullet point lists. (This does currently work in `master`)

> I read all of your commits so far and it mostly seemed to work for me, altough it's become quite more complex lol.

I wish I could come up with an easier way I really do :pensive: 

> Sure, would you rather I trim on this one or should I close this one and open a new one just to keep it clean?

Yeah I kinda spammed comments in this one lol. Let's close it and open new PRs for our separate changes.
Thank you for the burst of motivation to work on this issue!

G1org1owo:
Gotcha, I'll close this now, clean up on my end and open a new PR for the icons only.

> Thank you for the burst of motivation to work on this issue!

Thank you for the great work!

Issue #745, opened 2025-08-02
Recommendation: drop pyheif and heif-image-plugin in favor of pillow-heif and drop pillow-avif-plugin due to it being obsolete


kuflierl:
pyheif is an unmaintained library that has been unmaintained for a year and has been broken for 2 libheif releases already. pillow-heif can safely replace the combination of heif-image-plugin and pyheif.

pillow-avif-plugin has long been merged into upstream pillow and can now be dropped after an update to pillow.

pyheif has been [broken on nixpkgs/NixOS for a while](https://github.com/NixOS/nixpkgs/issues/400064) and can not really be fixed as long as upstream is unmaintained.

PR #751, opened 2025-09-18
server: replace pyheif with pillow-heif, bump alpine version to v3.22

kuflierl:
- Bump alpine to version v3.22
- Replaces pyheif with pillow-heif (pyheif is currently unmaintained upstream and broken with current libheif version)
- removes pillow-avif-plugin requirement due to upstream pillow support

Related to:
- https://github.com/NixOS/nixpkgs/pull/443257
- #745

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,4 +1,4 @@
-ARG ALPINE_VERSION=3.13
+ARG ALPINE_VERSION=3.22
 
 
 FROM alpine:$ALPINE_VERSION as prereqs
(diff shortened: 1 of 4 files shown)


po5:
Thanks.
Can you explain the Alpine version bump? Is newer Python required for this new version of Pillow?
This obviously needs to be looked over carefully if we're to update the Python version, but I don't know if that's even what's happening so please clear this up for me.

kuflierl:
Alright, here is the rundown. I wanted to use a later version of pillow due to its native support for AVIF files. I also noticed that alpine 3.13 is chronically EOL (Oldest supported version is 3.19). The simplest and most secure way for me to get that latest pillow version was to get an up-to date version of Alpine so i went with that.

As far as i remember the python version isn't really a concern. Alpine 3.22 uses python 3.12.11 und Alpine 3.13 uses Python 3.8.15. This might sound like a giant version leap but as far as we can see on the Nixpkgs side, nothing broke even on python 3.13.

My main concern is that pyheif hasn't been updated to work with libheif since libheif version 1.19 (we are at 1.20)
(https://github.com/carsales/pyheif/issues/103).

PR #756, opened 2025-11-30
Don't load already uploaded files into memory

G1org1owo:
The current methods for post creation/update work using the full `content` as a byte array. When working with large files, such as videos, this can be a problem, as the data ends up being duplicated in a few occasions.

The most absurd one is having to create a temporary file for use with ffmpeg, when the original file could have been used all along without even loading it into the python process' memory.

This is a draft that addresses this issue by adding to those methods a new optional parameter, `content_file`, which is used if `content` is `None` or empty. The optimizations are as follows:

- Methods that need to access the actual contents, such as `mime.get_mime_type`, are implemented in a way that severely limits the total data being read (20 bytes).

- To efficiently implement `util.get_md5` and `util.get_sha1`, I have introduced `util.get_checksums_from_file`, which loads 2M at a time to update both the md5 and sha1 hashes.

- If `content_file` is set and `content` is `None` or empty, `__content_file` is set rather than `__content`, and `shutil.copyfile` is used to copy the original file into the correct data location. Another improvement here could have been using `os.rename`, but I was not sure I could operate under the assumption temp files obtained via API upload are always meant to be used only once. I'll gladly take suggestions on this one.

The parameter ordering for `posts.create_post` is quite weird, but I decided against breaking compatibility with the older signature rather than updating it everywhere as I'm not really sure if there are others who have made external scripts using it like me.

I'm open to any comments or suggestions, especially please tell me if you have a better name for `_execute_impl` as I'm not really a fan of this naming convention but the only other name I could think of in that moment was `_execute2`.

### DIFF
--- a/server/szurubooru/func/file_uploads.py
+++ b/server/szurubooru/func/file_uploads.py
@@ -22,6 +22,10 @@ def get(checksum: str) -> Optional[bytes]:
     return files.get("temporary-uploads/%s.dat" % checksum)
 
 
+def get_handle(checksum: str) -> Optional[BufferedIOBase]:
+    return files.get_handle("temporary-uploads/%s.dat" % checksum)
+
+
 def save(content: bytes) -> str:
     checksum = util.get_sha1(content)
     path = get_path(checksum)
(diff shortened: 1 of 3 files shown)


po5:
> Another improvement here could have been using `os.rename`, but I was not sure I could operate under the assumption temp files obtained via API upload are always meant to be used only once. I'll gladly take suggestions on this one.

I think we should keep the current behavior just like you did. Stale files will automatically get removed. I guess it kind of sucks for non-CoW filesystems though.

PR #770, opened 2026-05-20
server/net: support outbound socks/http proxies

starkalexane:
If you put (for example) socks5h://localhost:9050 as the proxy,
all outbound requests (yt-dlp and direct fetch) will use Tor. Optionally, the webhook requests will use Tor, but this is optional because the webhook URLs are configured statically by the admin, so it's not always necessary that they be proxied.

To *properly* fix SSRF vulnerabilities (supposing a proxy allows LAN requests), one would need something more advanced with IP address filtering (as [Matrix Synapse](https://github.com/element-hq/synapse) does). Alternatively, a filter can be placed on the entire python process or something (as [0x0.st](https://git.0x0.st/mia/0x0) does, suggesting [FireHOL](https://firehol.org/)).

I tested this code using Tor for upload, webhook, and yt-dlp. I did not test HTTP proxies, but I would be very surprised if yt-dlp supported SOCKS5h proxies and *not* HTTP proxies lol. I also did not run the included tests because I'm not sure how to do it from my OS. If you need me to, I can do it in about a week.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -30,7 +30,8 @@ RUN pip3 install --no-cache-dir --disable-pip-version-check \
         "pyheif==0.6.1" \
         "heif-image-plugin>=0.3.2" \
         yt-dlp \
-        "pillow-avif-plugin~=1.1.0"
+        "pillow-avif-plugin~=1.1.0" \
+        "requests[socks]>=2.34.2,<3.0.0"
 RUN apk --no-cache del py3-pip
 
 COPY ./ /opt/app/
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -31,7 +31,7 @@ RUN pip3 install --no-cache-dir --disable-pip-version-check \
         "heif-image-plugin>=0.3.2" \
         yt-dlp \
         "pillow-avif-plugin~=1.1.0" \
-        "requests[socks]>=2.34.2,<3.0.0"
+        "requests[socks]>=2.32.4,<3.0.0"
 RUN apk --no-cache del py3-pip
 
 COPY ./ /opt/app/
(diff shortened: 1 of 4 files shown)


po5:
Thank you, this looks good! It bothered me too that fetch was done in clear.
My only objection is that we should not require `requests>=2.34.2`, because it bumps the required Python version all the way up to 3.10.
I think the newest `request` version we can *require* is 2.27.1, because all other files in the project run on 3.6.
Does that seem acceptable?

I think that `server/Dockerfile` should also be updated to install `py3-requests`.

starkalexane:
> I think the newest `request`[sic] version we can _require_ is 2.27.1, because all other files in the project run on 3.6.

3.6 eh?

https://github.com/rr-/szurubooru/blob/ada1a81260b4a84cea87d28b6904f1a27e3eef3b/server/szurubooru/func/net.py#L42

This requires 3.8. The latest `requests` version that supports 3.8 is 2.32.4.

I will update the Dockerfile.

starkalexane:
I don't think I can put py3-requests in the Dockerfile, because we also need the `[socks]` extra to support `socks5h://` for Tor.

po5:
Will test the Docker setup when I can, thanks. Sorry if it takes me a while.

---

About 3.6: Yes, I was very intentional in that comment. Every file outside of net.py works in Python 3.6, and I'm not happy about the few commits that bumped the version requirement for that file alone for no actual gain.

Keeping it at 3.8 is honestly fine because we've had it for a while and won't be able to go any lower when we switch away from [doing multipart handling with the cgi module](https://github.com/rr-/szurubooru/pull/756#issuecomment-3616959716), and use [multipart](https://pypi.org/project/multipart/) instead.

PR #772, opened 2026-06-24
Added OIDC Support

Lachee:
Adds OIDC Support for login

Configurable with the server config:
```yaml
# OpenID Connect (OIDC) single sign-on
# Set client_id, client_secret, and issuer to enable SSO login.
# redirect_uri must be your site's base URL (the SPA root), registered with
# your OIDC provider exactly as written here.
oidc:
    client_id:       # example: my-client
    client_secret:   # example: mysecret
    issuer:          # example: https://auth.example.com
    # must be registered with your OIDC provider and must point to your site root
    redirect_uri:    # example: https://example.com
    scopes: 'openid email profile'
    # set to no to prevent automatic account creation on first OIDC login
    allow_registration: yes
    # disable username/password login and self-registration; users must log in via SSO
    disable_password_auth: no
    # optional custom label for the SSO button (default: 'Log in with SSO')
    button_label:
    # optional URL to an icon image shown on the SSO button
    button_icon:
    # OIDC claim containing the user's avatar/profile picture URL
    # standard value is 'picture'; leave blank to disable avatar import
    avatar_claim: picture
 ```
 
 You can:
- Setup OpenID Connect
- Disable password logins
- Disable new accounts
- Customise login button
- Auto-load a users avatar.

In the future, there could be a potential for roles based on the OIDC claims too, such as with PocketID's groups.

### DIFF
--- a/client/css/login-view.styl
+++ b/client/css/login-view.styl
@@ -1,0 +1,12 @@
+#login
+    .oidc-icon
+        height: 2lh
+        width: auto
+        vertical-align: middle
+        margin-right: 0.4em
+
+    .oidc-login-btn
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        gap: inherit;
--- a/client/html/login.tpl
+++ b/client/html/login.tpl
@@ -1,36 +1,54 @@
 <div class='content-wrapper' id='login'>
     <h1>Log in</h1>
-    <form>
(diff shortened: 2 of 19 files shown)


Lachee:
I will be honest, I didnt mean to make a PR to the main repo, but it's here now so might as well push it forward to the community.

PR #775, opened 2026-07-01
Add Smilingwolf wd-tagger as an automatic tagger for posts

Dominik7272:
This PR adds an automatic image tagger using any [SmilingWolf](https://huggingface.co/SmilingWolf) model.
The tagger is deployed as a docker container.

## Features
- Automatically tags every image upon upload
- Safety classification (safe, sketchy, unsafe)
- Tags will be put into categories (general, character)
- Manual per-image tagging in the post edit menu
- Supports uncommon image formats like AVIF

All settings including what model to use are inside ./server/config.yaml.

### DIFF
--- a/client/css/tag-input-control.styl
+++ b/client/css/tag-input-control.styl
@@ -159,3 +159,7 @@ div.tag-input, ul.compact-tags
     div.tag-input, ul.compact-tags
         .tag-usages, .tag-weight, .remove-tag
             color: $inactive-link-color-darktheme
+
+.auto-tag-btn
+    width: 100%
+    margin-top: 0.5em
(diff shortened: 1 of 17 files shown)


Lachee:
well this is interesting! 

I had [made a solution](https://github.com/Lachee/szurubooru-autotag) for my instance using [Taggerine](https://huggingface.co/Hetaneko/taggerine) and using webhooks to link it and the existing szurubooru API. How does this one compare to something like Taggerine?

I think, since this is a field with constant growth, it might be better to have a way for instances to define their own tagging solutions ? Maybe a common API that the server can callback to?

Particularly, hosts might want to put the AI models on another machine with GPU acceleration, or even  not use AI at all for ethical reasons. It also adds quite a bit of size to the compose, which while editable, might not be a good "default". ( Also note that MPS acceleration for Apple Sillicon isn't available with Docker. Basically only Cuda. )

